PR #6, opened 2020-07-30
Add github actions to run lint and build

vaniisgh:
setup a build action
some left overs from renaming also covered.

ref to running action with the same branch: https://github.com/vaniisgh/openebsctl/pull/2


.github/workflows/go.yml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14.4
+      id: go_1
+
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: run golint
+      run: make golint
```
prateekpandey14, line 37: We can add the verify deps as well ?

vaniisgh: yes, I think that would be good too, will just add it :)

PR #7, opened 2020-08-03
Update client

vaniisgh:
Update Client functions : 
- [x] use pass namespace to functions individually
- [ ] update GetCStorVolumeInfoMap
- [ ] change `attached` status optput in the volume-list command


kmova:
Will be taken up in a new PR.

PR #12, opened 2021-05-06
feat(cli) Re-order subcommands to `APP VERB NOUN`

vharsh:
## Summary of changes

- drop 'cstor', the castype from subcommand(in further PRs the sub-command will be inferred and listed as per plans and/or filtered by `--flag`, i.e. the `adjective`)
- reordering of sub-commands

## Need to improve in upcoming PRs

- passing
- need to refactor flag placements, etc

Benefits:
- feels similar to kubectl in use
- easier to add more resources and operation

Newer CLI hierarchy:
image
 

## User visible change

The ordering of commands change:
`kubectl openebs cstor volume list` becomes `kubectl openebs get volumes` and `kubectl openebs cstor volume info` becomes `kubectl openebs describe volume`; the output & logic for fetching remains same.

E.g.
```bash
# Before
dev@localhost:~$ kubectl cstor openebs pool list
Name              Namespace  HostName                       Free    Capacity   ReadOnly  ProvisionedReplicas  HealthyReplicas  Status  Age
----              ---------  --------                       ----    --------   --------  -------------------  ---------------  ------  ---
unfmt-cstor-kc7m  test       director-dev-cluster-1-node-1  48200M  48201580k  false     0                    0                ONLINE  1d10h


# After
dev@localhost:~$ kubectl openebs get pools
Name              Namespace  HostName                       Free    Capacity   ReadOnly  ProvisionedReplicas  HealthyReplicas  Status  Age
----              ---------  --------                       ----    --------   --------  -------------------  ---------------  ------  ---
unfmt-cstor-kc7m  test       director-dev-cluster-1-node-1  48200M  48201580k  false     0                    0                ONLINE  1d10h

```


kubectl-openebs/cli/command/get/pool_list.go
```diff
@@ -87,8 +85,6 @@ func RunPoolsList(cmd *cobra.Command) error {
 		fmt.Println("No Pools are found")
```
vharsh, line 85: Need to decide to use one definitive way of the logging errors/warnings, klog or fmt.Println.

Please vote
❤️  for klog
🚀  for fmt.Println

kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -14,10 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package volume
+package get
```
vharsh, line 17: I've made a mess with packaging right now, I sort of copied how things were in kubectl and created a get package with a get.go, after this I'm a bit confused w.r.t should there be a single `volume_cmd.go` which will have all functionality talking to the commands & flags about `get`, `describe`, `delete`, `watch` (might get added later if useful).

kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -32,39 +31,28 @@ This command displays status of available zfs Volumes.
...
 		Run: func(cmd *cobra.Command, args []string) {
 			util.CheckErr(RunVolumesList(cmd), util.Fatal)
 		},
 	}
-
-	cmd.Flags().StringVarP(&namespace, "namespace", "n", "openebs",
-		"namespace name, required if volume is not in the `default` namespace")
-
-	flag.CommandLine.Parse([]string{})
-
```
vharsh, line 58: This namespace flag, among with other nice flags from kubectl get included via the k8s.io/cli-runtime package 's genericclioptions, thus this is not required, that being said, a nicer way to move such info is still required so that this isn't done at each sub-command.

kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -32,39 +31,28 @@ This command displays status of available zfs Volumes.
...
 	return cmd
 }
 
-//RunVolumesList fetchs & lists the volumes
+// RunVolumesList lists the volumes
 func RunVolumesList(cmd *cobra.Command) error {
-
-	client, err := client.NewK8sClient(namespace)
+	// TODO: Why is this working?
```
vharsh, line 53: I tried this by installing openebs on the `test` namespace and this still worked, I believe `""` used to be the k8s constant for default or all-namespaces(my memory is a lot hazy here).

kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -32,39 +31,28 @@ This command displays status of available zfs Volumes.
...
+	client, err := client.NewK8sClient("")
 	util.CheckErr(err, util.Fatal)
```
vharsh, line 55: I tried this by installing openebs on the test namespace and this still worked, I believe "" used to be the k8s constant for default or all-namespaces(my memory is a lot hazy here).

kubectl-openebs/cli/command/get/pool_list.go
ParthS007: I looked into the [cobra library examples](https://github.com/spf13/cobra#using-the-cobra-library) and since we are following that, we can go with `fmt.PrintIn`.

kubectl-openebs/cli/command/get/volume_list.go
ParthS007: We can have a single file for now and as we move further, we can expand(move) the commands to different files. wdyt?

kubectl-openebs/cli/command/get/volume_list.go
ParthS007: if it working with `namespace` then I think it's good to keep that.

I am assuming `namespace` is an arg here provided by the cli user?

ParthS007: We can have a universal function to return the `client` using the `namepsace` as I see 
`client, err := client.NewK8sClient()` is repeated a couple of times. wdyt?

kubectl-openebs/cli/command/commands.go
```diff
@@ -17,26 +17,37 @@ limitations under the License.
 package command
 
 import (
-	"github.com/openebs/openebsctl/kubectl-openebs/cli/command/cstor"
+	"flag"
+
```
ParthS007, line 21:
```suggestion

```

kubectl-openebs/cli/command/commands.go
```diff
@@ -17,26 +17,37 @@ limitations under the License.
...
 // NewOpenebsCommand creates the `openebs` command and its nested children.
 func NewOpenebsCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "kubectl-openebs",
-		Short: "OpenEBSctl is a a tool for interacting with OpenEBS storage components",
-		Long:  `OpenEBSctl is a kubectl plugin to interact with OpenEBS container Attached Storage components. `,
+		Use:   "openebs",
+		Short: "openebs is a a kubectl plugin for interacting with OpenEBS storage components",
+		Long: `Find out more about OpenEBS on https://docs.openebs.io/
+`,
```
ParthS007, line 36:
```suggestion
		Long: `Find out more about OpenEBS on https://docs.openebs.io/`,
```

client/k8s.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
ParthS007, line 8: what's the intention of adding white space here?

client/k8s.go
```diff
@@ -214,7 +214,7 @@ func (k K8sClient) GetCStorVolumeInfoMap(node string) (map[string]*util.Volume,
 		pv, err := k.GetPV(*i.Spec.Source.PersistentVolumeName)
 
 		if err != nil {
-			klog.Errorf("Failed to get PV", i.ObjectMeta.Name)
+			klog.Errorf("Failed to get PV %s", i.ObjectMeta.Name)
```
ParthS007, line 217: `fmt.printIn` here if we decide to go with it.

ParthS007 (commented):
Thanks for the PR. LGTM 🚀 and left some minor cosmetic comments.

kubectl-openebs/cli/command/get/volume_list.go
vharsh: This is a great idea, this will make writing tests easier if it's initialized in the `openebs` sub-command level and passed around to the sub-sub-commands which'd need them.

vharsh: > if it working with namespace then I think it's good to keep that.

Yeah, but then if it was working with the namespace, it should likely break without it. I'll try to understand how this is still working.

> I am assuming namespace is an arg here provided by the cli user?

Yes, the flags nee to be defined at a command level, I saw they were defined multiple times(locally) so I created a PersistentFlag and then removed it because it's likely already coming from `kubectl-openebs/cli/command/commands.go`  line#40 in my PR.

```go
	kubernetesConfigFlags := genericclioptions.NewConfigFlags(true)
	kubernetesConfigFlags.AddFlags(cmd.PersistentFlags())
```

client/k8s.go
vharsh: I use vim to format my code before commit-ing(Goland doesn't run goimports,gofmt, etc correctly each time) and it gets rid of the space when I save a file. I'll drop this change.

client/k8s.go
vharsh: I had a small discussion with @Abhinandan-Purkait  and we both agree that`fmt.Println` looks better to show errors like `No volumes found`, `No pools found in XYZ namespace`, etc and `klog` looks better for developer facing logs like `timed-out reaching out the k8s-api`, etc. More finer discussion may be suggested by reviewers.

kubectl-openebs/cli/command/get/volume_list.go
vharsh: > We can have a universal function to return the client using the namespace as I see

I'll do that in upcoming PRs, this could be a big refactor, because I also need to support other quirks correctly(e.g. the end user passes the `-u` flag or `--kubeconfig` flag, etc. IMO, this should be initialized at the `openebs` command level and maybe passed around to other sub-command objects.

kubectl-openebs/cli/command/get/volume_list.go
vharsh: I am planning to move around logic and files in upcoming PRs, will discuss sometime, I'm sort of blindly following what k8s has already done^(this may not be the best approach) as we are solving a much smaller problem right now and may expand to a larger use-case.

^See the structure here -> https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kubectl/pkg/cmd the sig-cli probably went ahead with this structure because they've got a lot of sub-commands(like get, describe, explain, etc)

kmova (approved):

PR #19, opened 2021-05-25
List & describe objects by name

vharsh:
# What's changing?

- GetByObjectName support has been added; similar to `kubectl get po pod-name` which only shows one pod.
- Existing functionality without the object name doesn't change

# Summary of under-the-hood changes.

- the namespace column has been omitted from the `kubectl openebs get volumes` output as `PersistentVolume` is not a namespaced resource, that namespace is of the `CStorVolume` resource, it can be re-added if required.

# Minor changes

- typos in logs are fixed

# User visible change

```bash
# Before: All resources were listed/described
$ kubectl openebs get volumes pvc-abcd-1234
NAME
--------
pvc-defg-1234
pvc-pqrst...
pvc-abcd-1234

# After: Only the requested resources are listed/described
$ kubectl openebs get volumes pvc-abcd-1234
NAME
--------
pvc-abcd-1234
```

### DIFF
--- a/client/k8s.go
+++ b/client/k8s.go
@@ -142,11 +142,14 @@ func (k K8sClient) GetStorageClass(driver string) (*v1.StorageClass, error) {
 
 // GetCStorVolumeAttachment using the K8sClient's storage class client
 func (k K8sClient) GetCStorVolumeAttachment(volname string) (*cstorv1.CStorVolumeAttachment, error) {
-	vol, err := k.OpenebsCS.CstorV1().CStorVolumeAttachments("").Get(context.TODO(), volname, metav1.GetOptions{})
+	vol, err := k.OpenebsCS.CstorV1().CStorVolumeAttachments("").List(context.TODO(),
+		metav1.ListOptions{LabelSelector: fmt.Sprintf("Volname=%s", volname)})
 	if err != nil {
-		return nil, errors.Wrap(err, "error while while getting storage csi volume")
+		return nil, errors.Wrap(err, "Error from server (NotFound): CVA not found")
+	} else if vol == nil || len(vol.Items) == 0 {
+		return nil, fmt.Errorf("Error from server (NotFound): CVA not found for volume %s", volname)
 	}
-	return vol, nil
+	return &vol.Items[0], nil
 }
 
 // GetcStorVolumes using the K8sClient's storage class client
--- a/client/k8s.go
+++ b/client/k8s.go
@@ -244,7 +244,7 @@ func (k K8sClient) NodeForVolume(volName string) (string, error) {
 	return podInfo.Items[0].Spec.NodeName, nil
 }
 
-// GetcStorPools using the OpenEBS's Client
+// GetcStorPools returns a list CSPIs
 func (k K8sClient) GetcStorPools() (*cstorv1.CStorPoolInstanceList, error) {
 	cStorPools, err := k.OpenebsCS.CstorV1().CStorPoolInstances("").List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
@@ -355,6 +355,7 @@ func (k K8sClient) GetcStorPoolsByName(names []string) (*cstorv1.CStorPoolInstan
 	}, nil
 }
 
+// GetcStorVolumesByNames gets the CStorVolume resource from all namespaces
 func (k K8sClient) GetcStorVolumesByNames(vols []string) (*cstorv1.CStorVolumeList, error) {
 	cVols, err := k.OpenebsCS.CstorV1().CStorVolumes("").List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
--- a/client/k8s.go
+++ b/client/k8s.go
@@ -350,3 +350,25 @@ func (k K8sClient) GetcStorPoolsByName(names []string) (*cstorv1.CStorPoolInstan
 		Items: list,
 	}, nil
 }
+
+func (k K8sClient) GetcStorVolumesByNames(vols []string) (*cstorv1.CStorVolumeList, error) {
+	cVols, err := k.OpenebsCS.CstorV1().CStorVolumes("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "Error while while getting volumes")
+	}
+	csMap := make(map[string]cstorv1.CStorVolume)
+	for _, cv := range cVols.Items {
+		csMap[cv.Name] = cv
+	}
+	var list []cstorv1.CStorVolume
+	for _, name := range vols {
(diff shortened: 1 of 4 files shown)


client/k8s.go
```diff
@@ -249,7 +249,50 @@ func (k K8sClient) NodeForVolume(volName string) (string, error) {
...
+	poolMap := make(map[string]cstorv1.CStorPoolInstance)
+	for _, p := range cspi.Items {
+		poolMap[p.Name] = p
+	}
+	var list []cstorv1.CStorPoolInstance
+	for _, name := range names {
+		if pool, ok := poolMap[name]; ok {
+			list = append(list, pool)
+		} else {
+			klog.Errorf("Error from server (NotFound): pool %s not found", name)
```
Abhinandan-Purkait, line 270: We might want to use `fmt.Println()` as this would be a user facing text.

client/k8s.go
```diff
@@ -249,7 +249,50 @@ func (k K8sClient) NodeForVolume(volName string) (string, error) {
...
+	csMap := make(map[string]cstorv1.CStorVolume)
+	for _, cv := range cVols.Items {
+		csMap[cv.Name] = cv
+	}
+	var list []cstorv1.CStorVolume
+	for _, name := range vols {
+		if pool, ok := csMap[name]; ok {
+			list = append(list, pool)
+		} else {
+			klog.Errorf("Error from server (NotFound): cStorVolume %s not found", name)
```
Abhinandan-Purkait, line 292: We might want to use `fmt.Println()` as this would be a user facing text.

kubectl-openebs/cli/command/describe/volume_info.go
```diff
@@ -99,112 +98,109 @@ func RunVolumeInfo(cmd *cobra.Command, vols []string, ns string) error {
...
+
+		//5. cStor Volume Replicas
+		cvrInfo, err := clientset.GetCVR(volName)
+		if err != nil {
+			fmt.Printf("failed to get cStor Volume Replicas for %s\n", volName)
+		}
+		cSPCLabel := cstortypes.CStorPoolClusterLabelKey
+		volume := util.VolumeInfo{
+			AccessMode:              util.AccessModeToString(pvInfo.Spec.AccessModes),
+			Capacity:                volumeInfo.Status.Capacity.String(),
```
Abhinandan-Purkait, line 137: - We would need error handling here, as we are not terminating code on not getting the resource in 108, 113, 118, 130. This would result in an nil pointer exception if the `err` is not nil and `pvcInfo` is nil, similarly for `cvcInfo, cvrInfo`

kubectl-openebs/cli/command/describe/volume_info.go
```diff
@@ -99,112 +98,109 @@ func RunVolumeInfo(cmd *cobra.Command, vols []string, ns string) error {
...
+			fmt.Printf("failed to get PV %s\n", volName)
+		}
+		//3. cStor Volume Config
+		cvcInfo, err := clientset.GetCVC(volName)
+		if err != nil {
+			fmt.Printf("failed to get cStor Volume config for %s", volName)
+		}
+
+		//4. Get Node for Target Pod from the openebs-ns
+		nodeName, err := clientset.NodeForVolume(volName)
```
Abhinandan-Purkait, line 124: - This logic would show the node name of the target pod which is not important for the user as it is an internal resource.
- We should display the `cva` ownerNodeID here, as that is relevant to the application using the volume.

kubectl-openebs/cli/command/describe/volume_info.go
```diff
@@ -99,112 +98,109 @@ func RunVolumeInfo(cmd *cobra.Command, vols []string, ns string) error {
...
+		portalInfo := util.PortalInfo{
+			IQN:            volumeInfo.Spec.Iqn,
+			VolumeName:     volumeInfo.Name,
+			Portal:         volumeInfo.Spec.TargetPortal,
+			TargetIP:       volumeInfo.Spec.TargetIP,
+			TargetNodeName: nodeName,
+		}
+
+		// Print the output for the portal status info
+		tmpl, err = template.New("PortalInfo").Parse(portalTemplate)
```
Abhinandan-Purkait, line 172: - We have a util method that takes a generic template and does this job, that could be used here.-
Ref: https://github.com/openebs/openebsctl/pull/17/files#diff-b771e541bd0847ee044ef3d69f9875f032d4e2c427c89bac3f82b179b862bcb9

kubectl-openebs/cli/command/describe/volume_info.go
vharsh: 👍🏼 I'll set it to `N/A` in case the CVA isn't found for this resource.

client/k8s.go
```diff
@@ -224,6 +227,10 @@ func (k K8sClient) GetCVR(name string) (*cstorv1.CStorVolumeReplicaList, error)
 	if err != nil {
 		return nil, errors.Wrapf(err, "error while getting cStor Volume Replica for volume %s", name)
 	}
+	if len(CStorVolumeReplicas.Items) == 0 {
+		// TODO: This came during rebase, this shouldn't be required
+		klog.Errorf("Error while getting cStor Volume Replica for  %s , no replicas found", name)
```
Abhinandan-Purkait, line 232: - Do we want to show this to user as a log? Maybe a fmt.Println would have been good.

kubectl-openebs/cli/command/describe/volume_info.go
```diff
@@ -99,112 +96,107 @@ func RunVolumeInfo(cmd *cobra.Command, vols []string, ns string) error {
...
+			fmt.Printf("failed to get CStorVolumeAttachments for %s\n", volName)
+		} else {
+			nodeName = node.Spec.Volume.OwnerNodeID
+		}
+
+		//5. cStor Volume Replicas
+		cvrInfo, err := clientset.GetCVR(volName)
+		if err != nil {
+			fmt.Printf("failed to get cStor Volume Replicas for %s\n", volName)
+			continue
```
Abhinandan-Purkait, line 138: - We maybe do not want to skip the volume describe if replicas are not found. Since the CV, CVC, CVA exist we can show the volume details and show no replicas found in this case.

Abhinandan-Purkait (commented):
Minor Comments.

client/k8s.go
vharsh: I'll move it to show info to `fmt.Printf` or via klog(but without the file info), the error should be printed to `stderr`, a simple `fmt.Println` would print it to `stdout`.

Abhinandan-Purkait (approved):
LGTM

ParthS007 (approved):
LGTM

PR #22, opened 2021-06-03
Restructure code repository to help add support of different storage engines faster

vharsh:
# Other benefits/goals

- de-couple `cobra.Command` & business logic & k8s-client so that they are in different packages(separated concerns) & are independently testable
- support for different cas-types can be quickly added, by just adding one more valid-flag value, one more implementation

# Directory structure

// WIP: Will add a `tree` command output or a box diagram here when it's finalized fully
```bash
cmd/
├── completion
│   └── completion.go
├── describe
│   ├── describe.go
│   ├── pool.go
│   ├── pvc.go
│   └── volume.go
├── get
│   ├── get.go
│   ├── pool.go
│   └── volume.go
└── openebs.go

# TODO: Remove get & describe packages inside cas-types; think about use-case where functionality across cas-types is totally different, e.g. get-pool of cstor vs get-pool of jiva vs get vg-group of lvm-localpv
pkg/
├── client
│   ├── jiva.go
│   └── k8s.go
├── cstor
│   ├── describe
│   │   ├── pool_info.go
│   │   ├── pvc_info.go
│   │   └── volume_info.go
│   └── get
│       ├── pool_list.go
│       └── volume_list.go
├── interface.go
├── jiva
│   ├── describe
│   └── get
│       └── volume.go
└── util
    ├── checks.go
    ├── checks_test.go
    ├── constant.go
    ├── error.go
    ├── formatters.go
    ├── k8s_utils.go
    ├── types.go
    ├── util.go
    └── util_test.go
```

# Documentation tasks

Write docs to help contributors to 
- add new sub-commands, suggest naming convention for flags, etc
- add newer implementations for doing the same sub-task multiple times for different cas-types/storage engines


cmd/describe/volume.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	volumeInfoCommandHelpText = `
+This command fetches information and status of the various
+aspects of a cStor Volume such as ISCSI, Controller, and Replica.
+
+#
+$ kubectl openebs describe [pool|volume] [name]
```
shubham14bajpai: This should be `$ kubectl openebs describe volume [name] -n [namespace]` from the above text right?

cmd/get/volume.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"github.com/openebs/openebsctl/pkg/util"
+	"github.com/openebs/openebsctl/pkg/volume"
+	"github.com/spf13/cobra"
+)
+
+var (
+	volumesListCommandHelpText = `
+This command displays status of available zfs Volumes.
```
shubham14bajpai: This should be cstor volumes or this is generics so openebs volumes?

cmd/get/volume.go
```diff
@@ -0,0 +1,49 @@
...
+If no volume ID is given, a list of all known volumes will be displayed.
+
+Usage: kubectl openebs cStor volume list [options]
```
shubham14bajpai: Also usage should be:
```
$ kubectl openebs get volumes [options]
```

shubham14bajpai (commented):
Few nit comments about the usage text

pkg/cstor/get/volume_list.go
```diff
@@ -19,44 +19,17 @@ package get
 import (
```
Abhinandan-Purkait: This file is not needed anymore, functionality has beeen moved to `pkg/volume`

pkg/volume/volume.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// CasList returns a list of functions by cas-types for volume listing
+func CasList() []func(*client.K8sClient, *corev1.PersistentVolumeList, string) ([]metav1.TableRow, error) {
+	// a good hack to implement immutable lists in Golang & also write tests for it
+	return []func(*client.K8sClient, *corev1.PersistentVolumeList, string) ([]metav1.TableRow, error){GetJiva, GetCStor}
+}
+
+// CasMap returns a map cas-types to functions for volume listing
+func CasMap() map[string]func(*client.K8sClient, *corev1.PersistentVolumeList, string) ([]metav1.TableRow, error) {
```
Abhinandan-Purkait, line 66: Awesome Logic !!

Abhinandan-Purkait (commented):
We would not be needing `cstor` pkg as it would make more sense on having packages based on  types like, volume, storage, disks as discussed. Will that change be taken in this PR ?

pkg/volume/volume.go
Abhinandan-Purkait: Adding volume listing for other cas-types got pretty easy with this change.

cmd/completion/completion.go
```diff
@@ -57,16 +60,8 @@ To configure your zsh shell to load completions for each session add to your zsh
...
-	if len(args) == 0 {
-		fmt.Println("error: Shell not specified.")
-		return
-	}
-	if len(args) > 1 {
-		fmt.Println("error: Too many arguments. Expected only the shell type.")
-		return
-	}
 	if args[0] == "bash" {
 		RunCompletionBash(out, cmd)
```
Abhinandan-Purkait: Can we not call, `cmd.GenBashCompletion(out)` here, this would do the job of `RunCompletionBash`

cmd/completion/completion.go
```diff
@@ -78,13 +73,17 @@ func RunCompletion(out io.Writer, cmd *cobra.Command, args []string) {
...
-	cmd.GenBashCompletion(out)
+	err := cmd.GenBashCompletion(out)
+	if err != nil {
+		klog.Error(err)
+	}
 }
 
-//RunCompletionZsh is used for the zsh shell
+// RunCompletionZsh is used for the zsh shell
 func RunCompletionZsh(out io.Writer, cmd *cobra.Command) {
```
Abhinandan-Purkait: Can we use, `cmd.GenZshCompletion(out)` in line:70, this would do the job of `RunCompletionZsh`

pkg/volume/volume.go
vharsh: I've written some basic tests which would fail the tests in case the number of volume-implementations in the list-by-object & list-all differ.

cmd/get/volume.go
vharsh: This is supposed to be generic, I've updated it.

vharsh:
> We would not be needing cstor pkg as it would make more sense on having packages based on types like, volume, storage, disks as discussed. Will that change be taken in this PR ?

I'll do that in next PRs, as restructuring of code for storage-pools/vg-groups & devices(BDs) will need more insight into requirements by cas-type & some more insights as to how separation of concerns can be achieved with minimal code & effort.

Abhinandan-Purkait (approved):
Thank you! LGTM

shubham14bajpai (approved):
LGTM

PR #27, opened 2021-06-10
[bug]handle exception when cva is not present.

Abhinandan-Purkait:


kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -79,16 +79,30 @@ func RunVolumesList(cmd *cobra.Command, openebsNs string, vols []string) error {
...
+				val.Node}})
+		} else {
+			rows = append(rows, metav1.TableRow{Cells: []interface{}{
+				item.ObjectMeta.Namespace,
+				item.ObjectMeta.Name,
+				item.Status.Phase,
+				item.VersionDetails.Status.Current,
+				util.ConvertToIBytes(item.Status.Capacity.String()),
+				"N/A",
+				"N/A",
```
vharsh, line 101: (low-priority) This can be set to `NotAttached`, need to check if Kubernetes has a term or a state for this.

kubectl-openebs/cli/command/get/volume_list.go
```diff
@@ -79,16 +79,30 @@ func RunVolumesList(cmd *cobra.Command, openebsNs string, vols []string) error {
...
-			item.ObjectMeta.Namespace,
-			item.ObjectMeta.Name,
-			item.Status.Phase,
-			item.VersionDetails.Status.Current,
-			util.ConvertToIBytes(item.Status.Capacity.String()),
-			pvols[item.ObjectMeta.Name].StorageClass,
-			pvols[item.ObjectMeta.Name].AttachementStatus,
-			pvols[item.ObjectMeta.Name].AccessMode,
-			pvols[item.ObjectMeta.Name].Node}})
+		if val, ok := pvols[item.ObjectMeta.Name]; ok {
```
vharsh, line 82: (optional) assign the values from val.StorageClass, val.AttachmentStatus to a variable scoped outside this if-else and then call `append` on the `rows`, this will reduce the lines one would have to read. The effect is minimal(barely 5-8lines less) so this comment is optional.

vharsh (approved):
lgtm

kubectl-openebs/cli/command/get/volume_list.go
Abhinandan-Purkait: - Yep, adding to constants.

kubectl-openebs/cli/command/get/volume_list.go
Abhinandan-Purkait: - I will add variables.

PR #29, opened 2021-06-14
fix(listing, describe): show message and terminate if cluster has no cstor resources

Abhinandan-Purkait:
Temporary fix for https://github.com/openebs/openebsctl/projects/1#card-63009009 till #22 is at place.

This terminates the command execution if `cstor-csi-controller` is not found.


vharsh (approved):
lgtm

PR #45, opened 2021-07-07
Add LVM LocalPV volume listing support

### LINKED ISSUE #33: Add support for LVM LocalPV
# Potential scope

_More details will be added here, as more info drops by_

Some more fields are getting added to the LVM resources -> @ openebs/lvm-localpv -> 115

- ~Add a sub-command for getting volume-group(vg), as squishing this into the pool sub-command doesn't make good UX sense.~
- [x] pool sub-command will be re-named to something generic like `storage`, so that it can mean a pool of resources(& not a pool as in a cstor pool or an LVM volume-group) as mentioned in this recent [talks](https://www.youtube.com/watch?v=Bit4GnANuLc)
- [x]  CSI -> Get CR structures from https://github.com/openebs/lvm-localpv
- [x] Listing LVM volumes
- [x] Listing pools/LVM-nodes
- [x] Describing LVM volumes 
- [x] Describing pools/LVM-nodes, etc

vharsh:
## Notable changes

- K8s API Downgrade
    - k8s.io/* to 0.20.0 so that `openebs/lvm-localpv` can be imported without errors
    - other libraries such as `openebs/zfs-localpv` can be imported without errors as most OpenEBS projects are using 0.20.0 versions of `k8s.io/*`
- Uses RESTClient to get a list/map of LVMVolume
- The CSI controller version label from the helm chart/YAML is used to get the volume version 

improvement: #33


pkg/volume/lvmlocalpv.go
```diff
@@ -0,0 +1,64 @@
...
+	if ctrlPod, err := c.GetCSIControllerPod("openebs-lvm-controller"); err == nil {
+		version = ctrlPod.Labels["openebs.io/version"]
+	} else {
+		version = "N/A"
+	}
+	for _, pv := range pvList.Items {
+		var attachedNode, customStatus, ns string
+		_, lvmVolMap, err := c.GetLVMvol(nil, util.Map, "", util.MapOptions{Key: util.Name})
+		if err != nil {
+			return nil, fmt.Errorf("failed to list ZFSVolumes")
```
Abhinandan-Purkait: `failed to list ZFSVolumes` ----> `failed to list LVMVolumes`

pkg/client/k8s.go
```diff
@@ -621,3 +622,70 @@ func (k K8sClient) GetJVs(volNames []string, rType util.ReturnType, labelSelecto
...
+	return nil, nil, errors.New("invalid return type")
+}
+
+// GetCSIControllerPod returns the CSI controller pod with a specific
+// openebs-component-name label key
+func (k K8sClient) GetCSIControllerPod(name string) (*corev1.Pod, error) {
+	if pods, err := k.K8sCS.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("openebs.io/component-name=%s", name),
+	}); err == nil && len(pods.Items) == 1 {
+		// TODO: Think about Evicted pods
```
Abhinandan-Purkait: Can we make use of `pod.Spec.Phase` here to cross check Running?

Abhinandan-Purkait: We might also need to consider this `evicted` condition while creating `NamespaceMap`

Abhinandan-Purkait (approved):
LGTM. Some Minor comments.

pkg/client/k8s.go
vharsh: I think this can be done via field-selectors, I'll confirm once.

vharsh: Tested it, I've added `status.phase=Running` for `GetOpenEBSNamespace` which is being used for PVC-describe & `GetOpenEBSNamespaceMap` which isn't being used anywhere. I've not added it in the method which gets the TargetPod objects. @shubham14bajpai

pkg/volume/lvmlocalpv.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/openebs/openebsctl/pkg/util"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// GetLVMLocalPV returns a list of LVM-LocalPV volumes
+func GetLVMLocalPV(c *client.K8sClient, pvList *corev1.PersistentVolumeList, openebsNS string) ([]metav1.TableRow, error) {
+	var rows []metav1.TableRow
+	var version string
+	if CSIctrl, err := c.GetCSIControllerPod("openebs-lvm-controller"); err == nil {
```
shubham14bajpai: Nice :+1:

pkg/client/k8s.go
shubham14bajpai: Instead of pod can we get sts with the label? The sts will have the latest applied configurations and pods can differ. Even if the controller is down to 0 pods the volume will be working fine and the version will remain same as sts.

shubham14bajpai (commented):
Small comment about sts over pod

pkg/volume/lvmlocalpv.go
vharsh: > Instead of pod can we get sts with the label? The sts will have the latest applied configurations and pods can differ. Even if the controller is down to 0 pods the volume will be working fine and the version will remain same as sts.


Are all CSI Controllers STS always?

pkg/client/k8s.go
vharsh: > Instead of pod can we get sts with the label?

Are all OpenEBS CSI Controllers STS always? If yes, then that'd be a one solution for all.

pkg/volume/lvmlocalpv.go
shubham14bajpai: Yes

vharsh: This is the closest one can come to volume/pool(VG) versions for storage entities which lack a data plane, not sure how relevant would this be for debugging purposes.

shubham14bajpai (approved):
LGTM

PR #47, opened 2021-07-07
Add ZFS LocalPV volume listing support

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

TODO: Update Description


pkg/client/k8s.go
```diff
@@ -621,3 +624,71 @@ func (k K8sClient) GetJVs(volNames []string, rType util.ReturnType, labelSelecto
 	}
 	return nil, nil, errors.New("invalid return type")
 }
+
+// GetZFSVols returns a list or a map of ZFSVolume depending upon rType & options
+func (k K8sClient) GetZFSVols(volNames []string, rType util.ReturnType, labelSelector string, options util.MapOptions) (*zfs.ZFSVolumeList, map[string]zfs.ZFSVolume, error) {
+	config := os.Getenv("KUBECONFIG")
```
vharsh, line 630: (for much later) (nitpick) Can initializing config be avoided here & `K8sClient.K8sCS` be used instead, via some utility methods like `BuildFromConfig`, etc, as this method is taking `k` as implicit argument but it's not using it.

I understand that the other option for this is going back to RESTclient, is a worse approach from many angles, we'd need to brainstorm after the release. I believe the current approach(your implementation) is the most stable & simple way forward for now.

go.mod
```diff
@@ -6,13 +6,18 @@ require (
...
 	github.com/pkg/errors v0.9.1
 	github.com/ryanuber/columnize v2.1.0+incompatible
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.0
-	k8s.io/api v0.21.0
-	k8s.io/apimachinery v0.21.0
-	k8s.io/cli-runtime v0.21.0
-	k8s.io/client-go v0.21.0
+	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
+	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 // indirect
```
vharsh, line 15: These two dependencies might get trimmed with `go mod tidy`, please check once.

pkg/client/k8s.go
```diff
@@ -621,3 +624,71 @@ func (k K8sClient) GetJVs(volNames []string, rType util.ReturnType, labelSelecto
...
+		default:
+			return nil, nil, errors.New("invalid map options")
+		}
+	}
+	return nil, nil, errors.New("invalid return type")
+}
+
+// GetCSIControllerSTS returns the CSI controller sts with a specific
+// openebs-component-name label key
+func (k K8sClient) GetCSIControllerSTS(name string) (*appsv1.StatefulSet, error) {
```
vharsh, line 684: 👍🏼 I'll copy this to my PR #45

pkg/util/constant.go
```diff
@@ -41,6 +41,8 @@ const (
 	OpenEBSCasTypeKeySc = "cas-type"
 	// CstorCasType cas type name
 	CstorCasType = "cstor"
+	// ZFSCasType cas type name
+	ZFSCasType = "zfslocalpv"
```
vharsh, line 45: Do we need a confirmation with the naming of the cas-type flags or `zfslocalpv` is acceptable by everyone?

vharsh (approved):
lgtm

go.mod
vharsh: Please bump up your golang compiler to `1.16`, this will likely solve the go.mod & go.sum out-of-sync issue when `go mod XYZ` commands are run again locally.

PR #60, opened 2021-07-17
feat: add golangci-lint to actions

### LINKED ISSUE #51: Add go linting tools to CI
# Background

[GolangCI-lint](https://golangci-lint.run/) is a popular linter for go-projects which has support for a lot of popular linters like govet, golint, gofmt, etc, so it gets a bit easier than having to write individual shell scripts & complex Makefiles. It's super configurable so it's warnings can be silenced & it runs with similar toggles locally, i.e. the config file works well for the local development & CI.

It has a host of tools like golint, gofmt, govet and a bunch of unofficial tools like dupl, depguard, etc which should be added. Per the docs, using the [official golangci-lint action](https://github.com/marketplace/actions/run-golangci-lint) might turn out to be a better approach than adding it via the Makefile in some other action/job.

burntcarrot:
Fixes #51.

Changes:

- add golangci-lint as an action
- modify golint target in Makefile
- move license check to goheader

Signed-off-by: Aadhav Vignesh <aadhav.n1@gmail.com>

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,27 +1,6 @@
 name: Go
 on: [push, pull_request]
 jobs:
-
-  lint:
-    name: Golint
-    runs-on: ubuntu-latest
-    steps:
-
-    - name: Set up Go 1.16
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1.16
-      id: go_1
-
-    - name: Checkout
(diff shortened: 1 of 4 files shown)


vharsh (approved):
lgtm

vharsh:
Thanks for the PR @burntcarrot 😃🎉

PR #69, opened 2021-08-04
Add zfs pool & volume describe

### LINKED ISSUE #37: Add support for zfs-localPV
The pool listing for `zfs-pool` can be clubbed into the `kubectl openebs get pool --cas-type=zfspool`(but it'd lack the status field, which CStor field already has). As of now, we are not sure if zfs pool listing should go in the same sub-command(simpler UX of the CLI) or a new one(`kubectl openebs get zfspool`).

Features
- [x] Volume listing
- [x] Pool listing
- [x] Volume describe
- [x] Pool describe

vharsh:
## Describe ZFS & CStor storage with & without flags

```bash
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe s zfs-node1

zfs-node1 Details :

HOSTNAME        : zfs-node1
NAMESPACE       : openebs
NUMBER OF POOLS : 1
TOTAL FREE      : 32 GiB

# also works with the `openebs-namespace` flag
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe s zfs-node1 --cas-type=zfslocalpv

zfs-node1 Details :

HOSTNAME        : zfs-node1
NAMESPACE       : openebs
NUMBER OF POOLS : 1
TOTAL FREE      : 32 GiB

# CSTOR still works, even without the flags, because it'd brute-force each implementation till one of them don't return a non-nil error which would stop the execution
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe s default-cstor-disk-rhwj

default-cstor-disk-rhwj Details :
----------------
NAME             : default-cstor-disk-rhwj
HOSTNAME         : ip-10-0-6-143.ec2.internal
SIZE             : 90 GiB
FREE CAPACITY    : 73 GiB
READ ONLY STATUS : false
STATUS	         : OFFLINE
RAID TYPE        : stripe
....

# CSTOR with explicit flags also works
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe s default-cstor-disk-rhwj --cas-type=cstor

default-cstor-disk-rhwj Details :
----------------
NAME             : default-cstor-disk-rhwj
HOSTNAME         : ip-10-0-6-143.ec2.internal
SIZE             : 90 GiB
FREE CAPACITY    : 73 GiB
READ ONLY STATUS : false
STATUS	         : OFFLINE
RAID TYPE        : stripe
....

# wrongly given cas-types won't work and throw an apt error
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs get s --cas-type=dummy
cas-type dummy is not supported

```

## Describe volumes 

```bash
➜  openebsctl git:(zfs_desc) ✗ ./openebsctl describe v pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452

pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452 Details :
-----------------
Name          : pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452
Namespace     : openebs
AccessMode    : ReadWriteOnce 
CSIDriver     : zfs.csi.openebs.io
Capacity      : 4Gi
PVC           : csi-zfspv
VolumePhase   : Released
StorageClass  : openebs-zfspv
Version       : N/A
Status        : Ready
VolumeType    : DATASET
PoolName      : zfspv-pool
FileSystem    : zfs
Compression   : off
Deduplication : off
NodeID        : worker-sh1
Recordsize    : 4k

# no cas-type flags required for describing volumes as that info is estimated from the CSI.Driver string & StorageClass(if one CSI driver can provision n different types of cas-type
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe v pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452

pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452 Details :
-----------------
Name          : pvc-43fcbc72-a45a-49d5-9ec3-e383fcb91452
Namespace     : openebs
AccessMode    : ReadWriteOnce 
CSIDriver     : zfs.csi.openebs.io
Capacity      : 4Gi
PVC           : csi-zfspv
VolumePhase   : Released
StorageClass  : openebs-zfspv
Version       : N/A
Status        : Ready
VolumeType    : DATASET
PoolName      : zfspv-pool
FileSystem    : zfs
Compression   : off
Deduplication : off
NodeID        : worker-sh1
Recordsize    : 4k

# jiva volume describe without explicit flags works
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe vol pvc-478a8329-f02d-47e5-8288-0c28b582be25

pvc-478a8329-f02d-47e5-8288-0c28b582be25 Details :
-----------------
NAME            : pvc-478a8329-f02d-47e5-8288-0c28b582be25
ACCESS MODE     : ReadWriteOnce 
CSI DRIVER      : jiva.csi.openebs.io
STORAGE CLASS   : openebs-jiva-csi-sc
VOLUME PHASE    : Released
VERSION         : 2.9.0
JVP             : jivavolumepolicy
SIZE            : 4.0 GiB
STATUS          : RW
REPLICA COUNT	: 1


Portal Details :
------------------
IQN              :  iqn.2016-09.com.openebs.jiva:pvc-478a8329-f02d-47e5-8288-0c28b582be25
VOLUME NAME      :  pvc-478a8329-f02d-47e5-8288-0c28b582be25
TARGET NODE NAME :  minikube
PORTAL           :  10.101.211.252:3260


No replicas found for the JivaVolumepvc-478a8329-f02d-47e5-8288-0c28b582be25

# jiva volume describe also works when only OPENEBS_NAMESPACE is specified(i.e. estimating OPENEBS_NAMESPACE when cas-type isn't provided almost always works)
➜  openebsctl git:(zfs_desc) ✗ kubectl openebs describe v pvc-478a8329-f02d-47e5-8288-0c28b582be25 --openebs-namespace=jiva

pvc-478a8329-f02d-47e5-8288-0c28b582be25 Details :
-----------------
NAME            : pvc-478a8329-f02d-47e5-8288-0c28b582be25
ACCESS MODE     : ReadWriteOnce 
CSI DRIVER      : jiva.csi.openebs.io
STORAGE CLASS   : openebs-jiva-csi-sc
VOLUME PHASE    : Released
VERSION         : 2.9.0
JVP             : jivavolumepolicy
SIZE            : 4.0 GiB
STATUS          : RW
REPLICA COUNT	: 1


Portal Details :
------------------
IQN              :  iqn.2016-09.com.openebs.jiva:pvc-478a8329-f02d-47e5-8288-0c28b582be25
VOLUME NAME      :  pvc-478a8329-f02d-47e5-8288-0c28b582be25
TARGET NODE NAME :  minikube
PORTAL           :  10.101.211.252:3260

No replicas found for the JivaVolumepvc-478a8329-f02d-47e5-8288-0c28b582be25

```

fixes: #37

### DIFF
--- a/cmd/describe/volume.go
+++ b/cmd/describe/volume.go
@@ -36,7 +36,6 @@ Advanced: Override the auto-detected OPENEBS_NAMESPACE
 
 // NewCmdDescribeVolume displays OpenEBS Volume information.
 func NewCmdDescribeVolume() *cobra.Command {
-	var casType string
 	cmd := &cobra.Command{
 		Use:     "volume",
 		Aliases: []string{"volumes", "vol", "v"},
@@ -49,6 +48,5 @@ func NewCmdDescribeVolume() *cobra.Command {
 			util.CheckErr(volume.Describe(args, openebsNS), util.Fatal)
 		},
 	}
-	cmd.PersistentFlags().StringVarP(&casType, "cas-type", "", "", "the cas-type filter option for fetching resources")
 	return cmd
 }
--- a/cmd/describe/volume.go
+++ b/cmd/describe/volume.go
@@ -36,6 +36,7 @@ Advanced: Override the auto-detected OPENEBS_NAMESPACE
 
 // NewCmdDescribeVolume displays OpenEBS Volume information.
 func NewCmdDescribeVolume() *cobra.Command {
+	var casType string
 	cmd := &cobra.Command{
 		Use:     "volume",
 		Aliases: []string{"volumes", "vol", "v"},
@@ -48,5 +49,6 @@ func NewCmdDescribeVolume() *cobra.Command {
 			util.CheckErr(volume.Describe(args, openebsNS), util.Fatal)
 		},
 	}
+	cmd.PersistentFlags().StringVarP(&casType, "cas-type", "", "", "the cas-type filter option for fetching resources")
 	return cmd
 }
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -554,7 +554,8 @@ func (k K8sClient) GetCSPIs(cspiNames []string, labelselector string) (*cstorv1.
 		if pool, ok := poolMap[name]; ok {
 			list = append(list, pool)
 		} else {
-			fmt.Printf("Error from server (NotFound): pool %s not found\n", name)
+			// This logging might be omitted
+			// fmt.Fprintf(os.Stderr, "Error from server (NotFound): pool %s not found\n", name)
 		}
 	}
 	return &cstorv1.CStorPoolInstanceList{
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -553,10 +553,11 @@ func (k K8sClient) GetCSPIs(cspiNames []string, labelselector string) (*cstorv1.
 	for _, name := range cspiNames {
 		if pool, ok := poolMap[name]; ok {
 			list = append(list, pool)
-		} else {
-			// This logging might be omitted
-			// fmt.Fprintf(os.Stderr, "Error from server (NotFound): pool %s not found\n", name)
 		}
+		// else {
+		// This logging might be omitted
+		// fmt.Fprintf(os.Stderr, "Error from server (NotFound): pool %s not found\n", name)
+		//}
 	}
 	return &cstorv1.CStorPoolInstanceList{
 		Items: list,
--- a/pkg/client/zfslocalpv.go
+++ b/pkg/client/zfslocalpv.go
@@ -91,6 +91,50 @@ func (k K8sClient) GetZFSVols(volNames []string, rType util.ReturnType, labelSel
 }
 
 // GetZFSNodes return a list of ZFSNodes
-func (k K8sClient) GetZFSNodes() (*zfs.ZFSNodeList, error) {
-	return k.ZFCS.ZfsV1().ZFSNodes("").List(context.TODO(), metav1.ListOptions{})
+func (k K8sClient) GetZFSNodes(volNames []string, rType util.ReturnType, labelSelector string, options util.MapOptions) (*zfs.ZFSNodeList, map[string]zfs.ZFSNode, error) {
+	zfsNode, err := k.ZFCS.ZfsV1().ZFSNodes("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, nil, err
+	}
+	var list []zfs.ZFSNode
+	if len(volNames) == 0 {
+		list = zfsNode.Items
+	} else {
+		zvsMap := make(map[string]zfs.ZFSNode)
+		for _, zn := range zfsNode.Items {
+			zvsMap[zn.Name] = zn
+		}
+		for _, name := range volNames {
+			if zv, ok := zvsMap[name]; ok {
+				list = append(list, zv)
+			} else {
+				fmt.Printf("Error from server (NotFound): zfsVolume %s not found\n", name)
+			}
+		}
+	}
+	if rType == util.List {
+		return &zfs.ZFSNodeList{
+			Items: list,
+		}, nil, nil
+	}
+	if rType == util.Map {
+		znMap := make(map[string]zfs.ZFSNode)
+		switch options.Key {
+		case util.Label:
+			for _, zn := range list {
+				if vol, ok := zn.Labels[options.LabelKey]; ok {
+					znMap[vol] = zn
+				}
+			}
+			return nil, znMap, nil
+		case util.Name:
+			for _, zn := range list {
+				znMap[zn.Name] = zn
+			}
+			return nil, znMap, nil
+		default:
+			return nil, nil, fmt.Errorf("invalid map options")
+		}
+	}
+	return nil, nil, fmt.Errorf("invalid return type")
 }
--- a/pkg/client/zfslocalpv.go
+++ b/pkg/client/zfslocalpv.go
@@ -108,6 +108,7 @@ func (k K8sClient) GetZFSNodes(volNames []string, rType util.ReturnType, labelSe
 			if zv, ok := zvsMap[name]; ok {
 				list = append(list, zv)
 			} else {
+				// This might be omitted
 				fmt.Printf("Error from server (NotFound): zfsVolume %s not found\n", name)
 			}
 		}
--- a/pkg/storage/cstor.go
+++ b/pkg/storage/cstor.go
@@ -70,11 +70,14 @@ func GetCstorPools(c *client.K8sClient, pools []string) ([]metav1.TableColumnDef
 
 // DescribeCstorPool method runs info command and make call to DisplayPoolInfo to display the results
 func DescribeCstorPool(c *client.K8sClient, poolName string) error {
-	poolInfo, err := c.GetCSPI(poolName)
+	pools, err := c.GetCSPIs([]string{poolName}, "")
 	if err != nil {
-		return errors.Wrap(err, "Error getting pool info")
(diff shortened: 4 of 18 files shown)


shubham14bajpai (approved):

pkg/client/k8s.go
```diff
@@ -553,9 +553,11 @@ func (k K8sClient) GetCSPIs(cspiNames []string, labelselector string) (*cstorv1.
 	for _, name := range cspiNames {
 		if pool, ok := poolMap[name]; ok {
 			list = append(list, pool)
-		} else {
-			fmt.Printf("Error from server (NotFound): pool %s not found\n", name)
 		}
+		// else {
```
vharsh, line 557: This writes a log to stdout/err when the storage is _guessed_ by brute force, which isn't very optimal for good CLI UX.
E.g. If a user is trying to describe an LVM volume, it's not optimal to see `Error from server, CStor pool lvm-vol1 not found`

pkg/client/zfslocalpv.go
```diff
@@ -91,6 +91,51 @@ func (k K8sClient) GetZFSVols(volNames []string, rType util.ReturnType, labelSel
...
+		zvsMap := make(map[string]zfs.ZFSNode)
+		for _, zn := range zfsNode.Items {
+			zvsMap[zn.Name] = zn
+		}
+		for _, name := range volNames {
+			if zv, ok := zvsMap[name]; ok {
+				list = append(list, zv)
+			} else {
+				// This might be omitted
+				fmt.Printf("Error from server (NotFound): zfsVolume %s not found\n", name)
```
vharsh, line 112: [Enhancement] This will basically require another round of refactors, it could be returning a slice of errors or appending like 2 non nil errors and returning it to the caller stating that out of 10 nodes 2 aren't found.

pkg/storage/storage.go
```diff
@@ -64,21 +63,41 @@ func CasList() []func(*client.K8sClient, []string) ([]metav1.TableColumnDefiniti
 func Describe(storages []string, openebsNs, casType string) error {
 	// 1. Create the clientset
 	k, _ := client.NewK8sClient(openebsNs)
-	// 2. Get the namespaces
+	// 2. Get the namespace
 	nsMap, _ := k.GetOpenEBSNamespaceMap()
```
vharsh, line 67: (good-first-issue): Implement custom cas-type to namespace mapping logic for entities which are based in `kube-system` namespace.

pkg/storage/testdata_test.go
```diff
@@ -247,3 +247,17 @@ var zfsNode2 = zfs.ZFSNode{
...
+		Kind:       "ZFSNode",
+		APIVersion: "zfs.openebs.io/v1",
+	},
+	ObjectMeta: metav1.ObjectMeta{
+		Name:      "node3",
+		Namespace: "zfs",
+		// OwnerReference: refers to the K8s-node where the zfs volume is created
+	},
+	Pools: []zfs.Pool{{Name: "zfs-pool1", UUID: "15423895941648453428", Free: resource.MustParse("33285828")},
+		{Name: "zfs-poolX", UUID: "15423895941648453426", Free: resource.MustParse("33285828Ki")}},
```
vharsh, line 262: This is a weird example of a zfs-pool where-in the former is likely the **size in bytes**, I've added this to ensure that the sum of total free storage size evaluates correctly.

cmd/describe/volume.go
```diff
@@ -48,5 +49,6 @@ func NewCmdDescribeVolume() *cobra.Command {
 			util.CheckErr(volume.Describe(args, openebsNS), util.Fatal)
 		},
 	}
+	cmd.PersistentFlags().StringVarP(&casType, "cas-type", "", "", "the cas-type filter option for fetching resources")
```
Abhinandan-Purkait: We won't be needing this flag. We can probably remove it if it doesnot break anything.

pkg/volume/volume.go
```diff
@@ -94,6 +94,19 @@ func Describe(vols []string, openebsNs string) error {
 		if openebsNs == "" {
 			if val, ok := nsMap[casType]; ok {
 				k.Ns = val
+			} else if casType == util.ZFSCasType {
+				// edge case: ZFS LocalPV resources are not created in kube-system
```
Abhinandan-Purkait, line 98: Is it because the controller is likely to be in `kube-system`?

pkg/util/k8s_utils.go
```diff
@@ -68,6 +68,11 @@ func GetCasTypeFromPV(v1PV *corev1.PersistentVolume) string {
 				return val
 			}
 		}
+		if v1PV.Spec.CSI != nil {
```
Abhinandan-Purkait, line 71: Thanks. With this we wont be needing the sc to determine the cas-type.

pkg/storage/cstor.go
```diff
@@ -72,11 +70,14 @@ func GetCstorPools(c *client.K8sClient, pools []string) ([]metav1.TableColumnDef
 
 // DescribeCstorPool method runs info command and make call to DisplayPoolInfo to display the results
 func DescribeCstorPool(c *client.K8sClient, poolName string) error {
-	poolInfo, err := c.GetCSPI(poolName)
+	pools, err := c.GetCSPIs([]string{poolName}, "")
```
Abhinandan-Purkait, line 73: Why do we need this change?

Abhinandan-Purkait (commented):
Thank you. We can probably make the `cas-type` flag mandatory for the storage describe to handle cases when `lvm` and `zfs` pool might have same name, also that would cut the brute force cost. 

We can also probably remove the `cas-type` flag from the volume describe as we are able to figure out the cas-type in case of a volume.

pkg/volume/volume.go
vharsh: Yes, that's making it a bit hard, but it can be likely fixed with a bunch of more if-elses which I'll send out in a new PR.

pkg/util/k8s_utils.go
vharsh: SC will be useful if the Local-XYZ CSI Provisioners get merged into a single one.

pkg/storage/cstor.go
vharsh: This one will work agnostic of the `c.Ns` value.

kmova (approved):

pkg/storage/storage.go
```diff
@@ -64,21 +63,41 @@ func CasList() []func(*client.K8sClient, []string) ([]metav1.TableColumnDefiniti
...
+	// 4. Brute-force run describe the storage by all cas-type functions
 	for _, storageName := range storages {
-		// 3. Describe the storage
-		if list, ok := CasDescribeMap()[util.CstorCasType]; ok {
-			err := list(k, storageName)
-			if err != nil {
-				return err
+		for _, work := range CasDescribeList() {
+			if err := work(k, storageName); err == nil {
+				continue
```
Abhinandan-Purkait: IMO. We should remove the continue and let the brute force logic describe all the pools having same name as the user has not specified any cas-type here.

For ex, I have one lvm pool named `awesomestorage` and a zfs pool also named `awesomestorage`.

Now the user enters, `kubectl openebs describe storage awesomestorage`, without specifying the cas type.
We should describe both the `awesomestorage`.

Abhinandan-Purkait (approved):
One comment. Thanks

pkg/storage/storage.go
Abhinandan-Purkait: The above would make more sense with the cas-type filtering we are providing.

vharsh: Makes sense, I'll drop the `continue`

PR #70, opened 2021-08-05
feat(volume debugger): Add feature for debugging cstor volume, add the status updates.

Abhinandan-Purkait:
### What does this PR do?
- This PR adds the `--debug` flag with the PVC Describe for initiating debugging.
- This adds the first module, i.e `CR statuses and Usage statuses`
- This adds Utilites for `available capacity` `string coloring.`

### Usage and Output
Screenshot from 2021-08-05 12-22-16

### DIFF
--- a/pkg/persistentvolumeclaim/debug_cstor.go
+++ b/pkg/persistentvolumeclaim/debug_cstor.go
@@ -1,0 +1,198 @@
+/*
+Copyright 2020-2021 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package persistentvolumeclaim
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
+
+	cstortypes "github.com/openebs/api/v2/pkg/apis/types"
+	"github.com/openebs/openebsctl/pkg/client"
+	"github.com/openebs/openebsctl/pkg/util"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/printers"
+)
+
+// DebugCstorVolumeClaim is used to debug a cstor volume by calling various modules
+func DebugCstorVolumeClaim(k *client.K8sClient, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume) error {
+	var cstorResources util.CstorVolumeResources
+	cstorResources.PVC, _ = k.GetPVC(pvc.Name, "default")
+	cstorResources.PV = pv
+	if pv != nil {
+		cv, _ := k.GetCV(pv.Name)
+		cstorResources.CV = cv
+		cvc, _ := k.GetCVC(pv.Name)
+		cstorResources.CVC = cvc
+		cva, _ := k.GetCVA(util.CVAVolnameKey + "=" + pv.Name)
+		cstorResources.CVA = cva
+		cvrs, _ := k.GetCVRs(cstortypes.PersistentVolumeLabelKey + "=" + pv.Name)
+		cstorResources.CVRs = cvrs
+	}
+	sc, _ := k.GetSC(*pvc.Spec.StorageClassName)
+	if sc != nil {
+		cspc, _ := k.GetCSPC(sc.Parameters["cstorPoolCluster"])
+		cstorResources.CSPC = cspc
+		if cspc != nil {
+			cspis, _ := k.GetCSPIs(nil, "openebs.io/cas-type=cstor,openebs.io/cstor-pool-cluster="+cspc.Name)
+			cstorResources.CSPIs = cspis
+			expectedBlockDevicesInPool := make(map[string]bool)
+
+			for _, pool := range cspc.Spec.Pools {
+				dataRaidGroups := pool.DataRaidGroups
+				for _, dataRaidGroup := range dataRaidGroups {
+					for _, bdName := range dataRaidGroup.GetBlockDevices() {
+						expectedBlockDevicesInPool[bdName] = false
+					}
+				}
+			}
+
+			var presentBlockDevicesInPool []string
+			for _, pool := range cspis.Items {
+				raidGroupsInPool := pool.GetAllRaidGroups()
+				for _, item := range raidGroupsInPool {
+					presentBlockDevicesInPool = append(presentBlockDevicesInPool, item.GetBlockDevices()...)
+				}
+			}
+
+			cstorResources.PresentBDs, _ = k.GetBDs(presentBlockDevicesInPool, "")
+			for _, item := range cstorResources.PresentBDs.Items {
+				if _, ok := expectedBlockDevicesInPool[item.Name]; ok {
+					expectedBlockDevicesInPool[item.Name] = true
+				}
+			}
+			cstorResources.ExpectedBDs = expectedBlockDevicesInPool
+			cstorResources.BDCs, _ = k.GetBDCs(nil, "openebs.io/cstor-pool-cluster="+cspc.Name)
+
+		}
+	}
+	_ = resourceStatus(cstorResources)
+	return nil
+}
+
+func resourceStatus(crs util.CstorVolumeResources) error {
+	var totalCapacity, usedCapacity, availableCapacity string
+	totalCapacity = util.ConvertToIBytes(crs.PVC.Spec.Resources.Requests.Storage().String())
+	usedCapacity = util.ConvertToIBytes(util.GetUsedCapacityFromCVR(crs.CVRs))
+	if usedCapacity != "" {
+		availableCapacity = util.GetAvailableCapacity(totalCapacity, usedCapacity)
+		percentage := util.GetUsedPercentage(totalCapacity, usedCapacity)
+		if percentage >= 80.00 {
+			availableCapacity = color.HiRedString(availableCapacity)
+		} else {
+			availableCapacity = color.HiGreenString(availableCapacity)
+		}
+	}
+	_, _ = fmt.Fprint(os.Stdout, "Volume Usage Stats:\n-------------------\n")
+
+	util.TablePrinter([]metav1.TableColumnDefinition{
+		{Name: "Total Capacity", Type: "string"},
+		{Name: "Used Capacity", Type: "string"},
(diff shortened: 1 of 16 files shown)


pkg/util/util.go
```diff
@@ -124,3 +127,29 @@ func ConvertToIBytes(value string) string {
 	}
 	return humanize.IBytes(iBytes)
 }
+
+// GetAvailableCapacity returns the available capacity irrespective of units
+func GetAvailableCapacity(total string, used string) string {
+	totalBytes, _ := units.RAMInBytes(total)
+	usedBytes, _ := units.RAMInBytes(used)
```
vharsh, line 133: (low-priority) This function may not evaluate to correct numbers if storage units are in powers of ten, as this takes the 1024 powers in the `parseSize` args

pkg/persistentvolumeclaim/debug_persistentvolumeclaim.go
```diff
@@ -0,0 +1,77 @@
+/*
```
vharsh: Please rename this file to `debug.go`, the name & package name are sort of repetitive.

kmova:
@Abhinandan-Purkait - what happens if PVC or other objects are in some kind of errored status and user runs the regular command without the "--debug" flag. Trying to see if "--debug" is implicit or really needs to be exposed. 

One thing with "--debug" is the various stages shown clearly. 

Just curious what would have happened if we had run on the above setup without "--debug" flag.

Abhinandan-Purkait:
> @Abhinandan-Purkait - what happens if PVC or other objects are in some kind of errored status and user runs the regular command without the "--debug" flag. Trying to see if "--debug" is implicit or really needs to be exposed.
> 
> One thing with "--debug" is the various stages shown clearly.
> 
> Just curious what would have happened if we had run on the above setup without "--debug" flag.

If the user enters the above command without the debug flag, the command will hit the pvc describe code and it will do a cstor pvc describe with all the available CRs that are present at the time of describing.
```
❯ kubectl openebs describe pvc mongo --openebs-namespace=openebs

mongo Details :
------------------
NAME             : mongo
NAMESPACE        : default
CAS TYPE         : cstor
BOUND VOLUME     : pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc
ATTACHED TO NODE : minikube-1
POOL             : default-cstor-disk
STORAGE CLASS    : common-storageclass
SIZE             : 20 GiB
USED             : 647 MiB
CV STATUS	 : Healthy
PV STATUS        : Released

Target Details :
----------------
NAMESPACE   NAME                                                              READY   STATUS    AGE      IP           NODE
openebs     pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc-target-7487cbc8bc5ttzl   3/3     Running   79d15h   172.17.0.9   minikube

Replica Details :
-----------------
NAME                                                               TOTAL     USED      STATUS    AGE
pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc-default-cstor-disk-dcrm   992 MiB   647 MiB   Healthy   79d16h
pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc-default-cstor-disk-fp6v   992 MiB   647 MiB   Healthy   79d16h
pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc-default-cstor-disk-rhwj   682 MiB   832 MiB   Offline   79d16h

Additional Details from CVC :
-----------------------------
NAME          : pvc-b84f60ae-3f26-4110-a85d-bce7ec00dacc
REPLICA COUNT : 3
POOL INFO     : [default-cstor-disk-dcrm default-cstor-disk-fp6v default-cstor-disk-rhwj]
VERSION       : 2.1.0-ee
UPGRADING     : true
```

Now if the user enters the same command with the `--debug` flag it will hit the debug code which is specific only to debgging, i.e, showing status, showing the lost BDs which I had mentioned while creating the cspc but they are no longer present, showing events, usage stats, all of these information. 
In case we try to debug a `non-cstor pvc` it gracefully errors out.
```
❯ kubectl openebs describe pvc jiva-csi-pvc --debug
Debugging is currently not supported for jiva Cas Type PVCs
```

The reason for putting the `--debug` flag and moving the debugging logic to that was that the describe itself is heavy on information and the debug is also not implemented for other cas-types.

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/openebs/openebsctl/pkg/util"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/printers"
+)
+
+// DebugCstorVolumeClaim is used to debug a cstor volume by calling various modules
+func DebugCstorVolumeClaim(k *client.K8sClient, pvc *corev1.PersistentVolumeClaim, pv *corev1.PersistentVolume) error {
+	var cstorResources util.CstorVolumeResources
+	cstorResources.PVC, _ = k.GetPVC(pvc.Name, "default")
```
vharsh: Was this supposed to be assigned to the pvc which has been received from args just like how pv is assigned to cstorResources from args? Also should we merge the `GetPVC` & `GetPVCs` method? The argument ordering is different in both, so it's confusing when using them.

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -0,0 +1,198 @@
...
+	cstorResources.PV = pv
+	if pv != nil {
+		cv, _ := k.GetCV(pv.Name)
```
vharsh, line 41: (good-first-issue) Skip swallowing error & print potential things which are off when CV, CVC, CVR, CVA are missing.

kmova (approved):

pkg/util/util_test.go
```diff
@@ -138,3 +138,77 @@ func TestConvertToIBytes(t *testing.T) {
...
+		{
+			"Test 2",
+			args{
+				total: "21.66GiB",
+				used:  "12.221GiB",
+			},
+			"9.439GiB",
+		},
+		{
+			"Test 2",
```
vharsh: Please rename it

pkg/persistentvolumeclaim/debug_persistentvolumeclaim.go
```diff
@@ -0,0 +1,77 @@
...
+	pvcList, err := k.GetPVCs(namespace, pvcs, "")
+	if len(pvcList.Items) == 0 || err != nil {
+		return errors.New("no pvcs found corresponding to the names")
+	}
+	// 2. Get the namespaces
+	nsMap, _ := k.GetOpenEBSNamespaceMap()
+	// 3. Range over the list of PVCs
+	for _, pvc := range pvcList.Items {
+		// 4. Fetch the storage class, used to get the cas-type
+		// TODO: Add cas-type label in every storage engine pv
```
vharsh:
```suggestion
```
This will likely not happen for CSI provisioned PVs

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -0,0 +1,185 @@
...
+		cspc, _ := k.GetCSPC(sc.Parameters["cstorPoolCluster"])
+		cstorResources.CSPC = cspc
+		if cspc != nil {
+			cspis, _ := k.GetCSPIs(nil, "openebs.io/cas-type=cstor,openebs.io/cstor-pool-cluster="+cspc.Name)
+			cstorResources.CSPIs = cspis
+			expectedBlockDevicesInPool := make(map[string]bool)
+			// This map contains the list of BDs we specified at the time of Pool Creation
+			for _, pool := range cspc.Spec.Pools {
+				dataRaidGroups := pool.DataRaidGroups
+				for _, dataRaidGroup := range dataRaidGroups {
```
vharsh, line 62: Is there a GetAllRaidGroups method here?

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -0,0 +1,185 @@
...
+				raidGroupsInPool := pool.GetAllRaidGroups()
+				for _, item := range raidGroupsInPool {
+					presentBlockDevicesInPool = append(presentBlockDevicesInPool, item.GetBlockDevices()...)
+				}
+			}
+
+			// Mark the present BDs are true.
+			cstorResources.PresentBDs, _ = k.GetBDs(presentBlockDevicesInPool, "")
+			for _, item := range cstorResources.PresentBDs.Items {
+				if _, ok := expectedBlockDevicesInPool[item.Name]; ok {
```
vharsh, line 80: [TODO Later] Need to think & discuss if they should be only set to true if that BlockDevice is in Active state(not claimState).

The BD can go in Inactive/Unknown state depending on various factors such as disk detached, NDM daemonset pod in CLBO state on that node.

vharsh (commented):
minor comments

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Yes, they should be set to true as at the time of applying cspc yaml I added some bds which were present then. Afterwards later if somehow bd got deleted or name got changed somehow, the bd mentioned in the cspc wont match the present bd. The above map does the matching of the BDs I mentioned in CSPCs and the ones that are present now. Now to answer the question whether Acitve to be picked or not, we wont because here we would show the status, i.e Active/Inactive of the currently present BDs.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: That would be handled in the specific modules. The reason for swallowing error is that in this module we would show the status of the whatever is present rest of the things would be shown as `Not found`.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Ohh, I forgot to remove this. This was added for testing. Thank you for pointing out.

pkg/util/util.go
Abhinandan-Purkait: In that case we would need a method that can identify whether the unit is to be Bytes or IBytes then assign a method accordingly for that. Would take this up in a separate PR.

pkg/persistentvolumeclaim/debug_persistentvolumeclaim.go
Abhinandan-Purkait: Done.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Also, only one one possible error can be got here i.e its not found, and in that case the CR struct will be nil. This would save some cyclomatic complexity.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: No, here we are parsing CSPC spec.

Abhinandan-Purkait: CSPC spec only has the things user has defined at the time of creation.

vharsh (approved):
lgtm, cool feature

PR #71, opened 2021-08-09
feat(volume debugger): Add module for PVC events display

### LINKED ISSUE #42: Can the `describe pvc` help to determine - why a cStor Volume is not ready?
- offline pools / replicas
- offline targets

```

Credits: Shubham
```

- [x] module for CVR events
- [x] all module integration

Abhinandan-Purkait:
### What does this PR do?
- This PR adds the module for fetching and displaying PVC events.
- This also includes some refactors.
- Replaces the `faith/color` package with own implementation.

### Output
1. If `events` occurred.
Screenshot from 2021-08-09 13-49-57

2. If `events` did not occur.
Screenshot from 2021-08-09 13-52-42

### DIFF
--- a/pkg/persistentvolumeclaim/debug_cstor.go
+++ b/pkg/persistentvolumeclaim/debug_cstor.go
@@ -18,7 +18,6 @@ package persistentvolumeclaim
 
 import (
 	"fmt"
-	"github.com/fatih/color"
 	"os"
 
 	cstortypes "github.com/openebs/api/v2/pkg/apis/types"
@@ -100,9 +99,9 @@ func resourceStatus(crs util.CstorVolumeResources) error {
 		availableCapacity = util.GetAvailableCapacity(totalCapacity, usedCapacity)
 		percentage := util.GetUsedPercentage(totalCapacity, usedCapacity)
 		if percentage >= 80.00 {
-			availableCapacity = color.HiRedString(availableCapacity)
+			availableCapacity = util.ColorText(availableCapacity, util.Red)
 		} else {
-			availableCapacity = color.HiGreenString(availableCapacity)
+			availableCapacity = util.ColorText(availableCapacity, util.Green)
 		}
 	}
 	// 3. Display the usage status
@@ -190,7 +189,7 @@ func displayPVCEvents(k client.K8sClient, crs util.CstorVolumeResources) error {
 	events, err := k.GetEvents(fmt.Sprintf("regarding.name=%s,regarding.kind=PersistentVolumeClaim", crs.PVC.Name))
 	// 3. Display the events
 	if len(events.Items) != 0 && err == nil {
-		_, _ = fmt.Fprint(os.Stdout, "\nChecking PVC Events:", color.HiRedString(fmt.Sprintf(" %s %d! ", util.UnicodeCross, len(events.Items))), "-------->\n")
+		_, _ = fmt.Fprint(os.Stdout, "\nChecking PVC Events:", util.ColorText(fmt.Sprintf(" %s %d! ", util.UnicodeCross, len(events.Items)), util.Red), "-------->\n")
 		var crStatusRows []metav1.TableRow
 		for _, event := range events.Items {
 			crStatusRows = append(crStatusRows, metav1.TableRow{Cells: []interface{}{event.Action, event.Reason, event.Note, util.ColorStringOnStatus(event.Type)}})
@@ -198,10 +197,9 @@ func displayPVCEvents(k client.K8sClient, crs util.CstorVolumeResources) error {
 		defer util.TablePrinter(util.EventsColumnDefinitions, crStatusRows, printers.PrintOptions{})
 		return nil
 	} else if len(events.Items) == 0 && err == nil {
-		_, _ = fmt.Fprint(os.Stdout, "\nChecking PVC Events:", color.HiGreenString(fmt.Sprintf(" %s %d! ", util.UnicodeCheck, len(events.Items))), "-------->\n")
+		_, _ = fmt.Fprint(os.Stdout, "\nChecking PVC Events:", util.ColorText(fmt.Sprintf(" %s %d! ", util.UnicodeCheck, len(events.Items)), util.Green), "-------->\n")
 		return nil
 	} else {
 		return err
 	}
-
 }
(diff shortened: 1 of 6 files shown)


pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -140,7 +139,7 @@ func resourceStatus(crs util.CstorVolumeResources) error {
 	// 4. Display the CRs statuses
 	util.TablePrinter(util.CstorVolumeCRStatusColumnDefinitions, crStatusRows, printers.PrintOptions{})
 
-	_, _ = fmt.Fprint(os.Stdout, "\nReplica Statuses:\n-------------------\n")
+	_, _ = fmt.Fprint(os.Stdout, "\nReplica Statuses:\n-----------------\n")
```
vharsh: Just use Println unless we need to write to stderr

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -183,3 +181,25 @@ func resourceStatus(crs util.CstorVolumeResources) error {
 
 	return nil
 }
+
+func displayPVCEvents(k client.K8sClient, crs util.CstorVolumeResources) error {
+	// 1. Set the namespace of the resource to the client
+	k.Ns = crs.PVC.Namespace
+	// 2. Fetch the events of the concerned PVC
+	events, err := k.GetEvents(fmt.Sprintf("regarding.name=%s,regarding.kind=PersistentVolumeClaim", crs.PVC.Name))
```
vharsh, line 198: Please add a comment stating that for core resources the Kind of the resource is set to `""`, this will help avoid future mistakes. This was a good TIL from our last discussion.

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -183,3 +181,25 @@ func resourceStatus(crs util.CstorVolumeResources) error {
...
+	// 3. Display the events
+	if len(events.Items) != 0 && err == nil {
```
vharsh: (nit) please flip the order of `&&` operation, even though `events` isn't nil even when err is non-nil

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -183,3 +181,25 @@ func resourceStatus(crs util.CstorVolumeResources) error {
 
 	return nil
 }
+
+func displayPVCEvents(k client.K8sClient, crs util.CstorVolumeResources) error {
```
vharsh, line 193: TODO: (later) This should be converted into a method and moved to `pkg/client` someday, I believe this was done because one cannot make it generic right now with kind missing.

pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -183,3 +181,25 @@ func resourceStatus(crs util.CstorVolumeResources) error {
...
+	// 2. Fetch the events of the concerned PVC
+	events, err := k.GetEvents(fmt.Sprintf("regarding.name=%s,regarding.kind=PersistentVolumeClaim", crs.PVC.Name))
+	// 3. Display the events
+	if len(events.Items) != 0 && err == nil {
+		_, _ = fmt.Fprint(os.Stdout, "\nChecking PVC Events:", util.ColorText(fmt.Sprintf(" %s %d! ", util.UnicodeCross, len(events.Items)), util.Red), "-------->\n")
+		var crStatusRows []metav1.TableRow
+		for _, event := range events.Items {
+			crStatusRows = append(crStatusRows, metav1.TableRow{Cells: []interface{}{event.Action, event.Reason, event.Note, util.ColorStringOnStatus(event.Type)}})
+		}
+		defer util.TablePrinter(util.EventsColumnDefinitions, crStatusRows, printers.PrintOptions{})
```
vharsh: Can defer be dropped, as this is the last condition anyway & it's not like a cleanup task,

Ref: [Blog on defer](https://blog.golang.org/defer-panic-and-recover)

pkg/persistentvolumeclaim/debug_cstor_test.go
```diff
@@ -66,3 +68,53 @@ func TestDebugCstorVolumeClaim(t *testing.T) {
 		})
 	}
 }
+
+func Test_displayPVCEvents(t *testing.T) {
+	k, _ := client.NewK8sClient("openebs")
```
vharsh, line 73: Create a fake-client for UT for each test-suite.

vharsh (changes requested):
minor change suggestions, please use `fmt.Println()` instead of `fmt.FPrintf(os.Stdout, "....\n")`

Avoid writing `\n`, things can turn out to be **wildly** different for different non-Linux platforms.

pkg/persistentvolumeclaim/debug_cstor_test.go
Abhinandan-Purkait: UTs are currently not written for this package, I will raise a separate PR with all units at the end of this feature. They are just added to pass the golangci for now.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Yes, this needs to be removed.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: This method wont be useful for any other volume/cas-type and thus made it unexported and kept in this package.

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Replaced with Println

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Added comments

pkg/persistentvolumeclaim/debug_cstor.go
Abhinandan-Purkait: Done.

pkg/persistentvolumeclaim/debug_cstor.go
vharsh: No I mean this function code would have to be copied multiple times to different places to be used, which isn't something desirable, I'll ask around if this problem is fix-able somehow by getting the kind correctly.

vharsh (approved):
lgtm

PR #74, opened 2021-08-11
feat(lvm vol describe): Add feature to describe lvm volumes

### LINKED ISSUE #33: Add support for LVM LocalPV

Abhinandan-Purkait:
### What does this PR do?
- This PR adds support to describe a `LVM` volume.
- This adds an utility to maintain one name for a cas.
- This adds unit tests for this feature.
- Some refactors

### Usage and Output

```
❯ kubectl openebs describe vol pvc-868e92a5-16ce-4919-91c2-a12a7210202f

pvc-868e92a5-16ce-4919-91c2-a12a7210202f Details :
------------------
Name            : pvc-868e92a5-16ce-4919-91c2-a12a7210202f
Namespace       : openebs
AccessMode      : ReadWriteOnce 
CSIDriver       : local.csi.openebs.io
Capacity        : 4Gi
PVC             : csi-lvmpv
VolumePhase     : Released
StorageClass    : openebs-lvmpv
Version         : 0.6.0
Status          : Ready
VolumeGroup     : lvmvg
Shared          : no
ThinProvisioned : no
NodeID          : node1-virtual-machine   

```

### DIFF
--- a/pkg/util/constant.go
+++ b/pkg/util/constant.go
@@ -31,8 +31,8 @@ const (
 	ZFSCasType = "zfslocalpv"
 	// JivaCasType is the cas type name for Jiva
 	JivaCasType = "jiva"
-	// LVMLocalPV is the cas type name for LocalPV-LVM
-	LVMLocalPV = "lvmlocalpv"
+	// LVMCasType cas type name
+	LVMCasType = "lvmlocalpv"
 	// Healthy cstor volume status
 	Healthy = "Healthy"
 	// StorageKey key present in pvc status.capacity
@@ -83,21 +83,21 @@ var (
 	CasTypeAndComponentNameMap = map[string]string{
 		CstorCasType: CStorCSIControllerLabelValue,
 		JivaCasType:  JivaCSIControllerLabelValue,
-		LVMLocalPV:   LVMLocalPVcsiControllerLabelValue,
+		LVMCasType:   LVMLocalPVcsiControllerLabelValue,
 	}
 	// ComponentNameToCasTypeMap is a reverse map of CasTypeAndComponentNameMap
 	// NOTE: Not including ZFSLocalPV as it'd break existing code
 	ComponentNameToCasTypeMap = map[string]string{
 		"openebs-cstor-csi-controller": CstorCasType,
 		"openebs-jiva-csi-controller":  JivaCasType,
-		"openebs-lvm-controller":       LVMLocalPV,
+		"openebs-lvm-controller":       LVMCasType,
 	}
 	// ProvsionerAndCasTypeMap stores the cas type name of the corresponding provisioner
 	ProvsionerAndCasTypeMap = map[string]string{
 		CStorCSIDriver: CstorCasType,
 		JivaCSIDriver:  JivaCasType,
 		// NOTE: In near future this might mean all local-pv volumes
-		LocalPVLVMCSIDriver: LVMLocalPV,
+		LocalPVLVMCSIDriver: LVMCasType,
 		ZFSCSIDriver:        ZFSCasType,
 	}
 	// CstorReplicaColumnDefinations stores the Table headers for CVR Details
--- a/pkg/util/constant.go
+++ b/pkg/util/constant.go
@@ -28,11 +28,11 @@ const (
 	// CstorCasType cas type name
 	CstorCasType = "cstor"
 	// ZFSCasType cas type name
-	ZFSCasType = "zfslocalpv"
+	ZFSCasType = "localpv-zfs"
 	// JivaCasType is the cas type name for Jiva
 	JivaCasType = "jiva"
 	// LVMCasType cas type name
-	LVMCasType = "lvmlocalpv"
+	LVMCasType = "localpv-lvm"
 	// Healthy cstor volume status
 	Healthy = "Healthy"
 	// StorageKey key present in pvc status.capacity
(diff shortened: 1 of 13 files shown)


pkg/util/constant.go
```diff
@@ -31,8 +31,8 @@ const (
 	ZFSCasType = "zfslocalpv"
 	// JivaCasType is the cas type name for Jiva
 	JivaCasType = "jiva"
-	// LVMLocalPV is the cas type name for LocalPV-LVM
-	LVMLocalPV = "lvmlocalpv"
+	// LVMCasType cas type name
+	LVMCasType = "lvmlocalpv"
```
vharsh: We can change the flag name and avoid the normalize function, as this is an alpha product now, we can make such name changes.

pkg/util/constant.go
```diff
@@ -83,21 +83,21 @@ var (
...
-		LVMLocalPV:   LVMLocalPVcsiControllerLabelValue,
+		LVMCasType:   LVMLocalPVcsiControllerLabelValue,
 	}
 	// ComponentNameToCasTypeMap is a reverse map of CasTypeAndComponentNameMap
 	// NOTE: Not including ZFSLocalPV as it'd break existing code
 	ComponentNameToCasTypeMap = map[string]string{
 		"openebs-cstor-csi-controller": CstorCasType,
 		"openebs-jiva-csi-controller":  JivaCasType,
-		"openebs-lvm-controller":       LVMLocalPV,
+		"openebs-lvm-controller":       LVMCasType,
```
vharsh: Use the named variable `LVMLocalPVcsiControllerLabelValue`.

pkg/util/util.go
```diff
@@ -171,3 +171,14 @@ func ColorStringOnStatus(stringToColor string) string {
 		return ColorText(stringToColor, Red)
 	}
 }
+
+// NormalizeCas is used to maintain one version of cas name
+func NormalizeCas(casType string) string {
```
vharsh: Lets get rid of this function if the flag value strings won't have trouble getting a `-` in the flag value, i.e. if `kubectl openebs get storage --cas-type=localpv-lvm` passes the cas-type to programs correctly we can rename the `lvmlocalpv` to `localpv-lvm`.

pkg/volume/lvmlocalpv.go
```diff
@@ -63,3 +82,44 @@ func GetLVMLocalPV(c *client.K8sClient, pvList *corev1.PersistentVolumeList, ope
...
 	return rows, nil
 }
+
+// DescribeLVMLocalPVs describes a single lvm-localpv volume
+func DescribeLVMLocalPVs(c *client.K8sClient, vol *corev1.PersistentVolume) error {
+	if vol == nil {
+		return fmt.Errorf("LVM volume nil")
+	}
+	lVols, _, err := c.GetLVMvol([]string{vol.Name}, util.List, "", util.MapOptions{})
+	if err != nil {
```
vharsh: Do add a check for the length of `lVols.Items`, the error can be non-nil and it would panic in the next line

pkg/volume/volume.go
```diff
@@ -94,20 +94,8 @@ func Describe(vols []string, openebsNs string) error {
...
-				}
-				if zfs != nil && zfs.Items != nil && len(zfs.Items) > 0 {
-					k.Ns = zfs.Items[0].Namespace
-				}
-				if err != nil {
-					return err
-				}
-			} else {
+			} else if casType != util.ZFSCasType && casType != util.LVMCasType{
+				// The reason for above condition is that, newer lvm has cas ty
```
vharsh, line 98: lvm & zfs aren't in the nsMap

vharsh (commented):
minor suggestions

pkg/util/constant.go
Abhinandan-Purkait: Changing the flag name surely helps. I did this way because the `zfs` is called as `zfslocalpv` thus to make things similar and avoid confusion.

pkg/util/util.go
Abhinandan-Purkait: Yes, we can but to maintain uniformity this has been added. Also normalizing stuffs would give users to give any thing and still get an output at an overhead of this matching and all. 

I was thinking of making the normalize recognize the cas-name from any string, viz `localpv-lvm`, `lvmlocalpv`, `lvm` whatever user enters by having some regex matching in here, does that sounds good to? 😅

pkg/volume/volume.go
Abhinandan-Purkait: - We donot really need them to be in the nsMap, as their controllers would be in `kube-system`. The above condition is to handle `cstor` and `jiva` ns not determined from nsMap

pkg/util/constant.go
vharsh: I'm not sure about the exact naming of it too, I created it as `lvmlocalpv` to avoid having a hyphen in the name.

pkg/util/util.go
vharsh: It sounds okay-ish, I don't know the exact name too 😓 

@shubham14bajpai @prateekpandey14 some naming help please

pkg/volume/volume.go
vharsh: got it, thanks

pkg/util/util.go
Abhinandan-Purkait: https://github.com/openebs/zfs-localpv/blob/95d5d3a8d33763add6df81a5829ba7800e71a132/pkg/zfs/volume.go#L64

https://github.com/openebs/lvm-localpv/blob/929ae4439f2da71a2d6ee5bda6a33dd2f7d424fc/pkg/lvm/volume.go#L59

Okay, I had added a hypen seperated name in both the repos. So I thing we should keep these in our code as well and get rid of the normalize method. If this sounds good.

vharsh: Yes, the localpv-lvm & localpv-zfs names seem better for the flags & internally.

On a nitpick side, things are a bit different in different places, but it'd likely converge to a single set of names sooner or later, I don't like having a normalize function everywhere.

```bash
(node-disk-manager) $ git grep StorageEngine

blockdevice/blockdevice.go:413: UsedBy StorageEngine
blockdevice/blockdevice.go:416:// StorageEngine is a typed string for the storage engine
blockdevice/blockdevice.go:417:type StorageEngine string
blockdevice/blockdevice.go:421: CStor StorageEngine = "cstor"
blockdevice/blockdevice.go:424: ZFSLocalPV StorageEngine = "zfs-localpv"
blockdevice/blockdevice.go:427: Mayastor StorageEngine = "mayastor"
blockdevice/blockdevice.go:430: LocalPV StorageEngine = "localpv"
blockdevice/blockdevice.go:433: Jiva StorageEngine = "jiva"
```

Issue #72, opened 2021-08-10
Make OpenEBS CLI easier to install via krew


vharsh:
Add the OpenEBS CLI to the krew index github repo or setup a custom krew index

- [x] Decide the repo for krew index, viz: custom or default(raise a PR)
- [x] Add support for the kubectl-auth-plugin for localpv-{lvm,zfs} 
- [x] Add LICENSE file to the 0.3 release bundle & ensure krew plugin spec for the plugin is updated via goreleaser
- [x] Implement that approach & it's pre-requisites & raise a PR to the default krew index(WIP)
- [x] Update the installation shell script with the newer path of the binary
- [x] Update README.md with krew installation steps
- [x] Automatically update the krew index whenever a new release is pushed, #87

PR #76, opened 2021-08-12
Minor code & config fixes for inclusion to the krew index

vharsh:
Summary of changes:

1. Add auth plugins for localpv clientsets, ref: [auth-plugins](https://github.com/kubernetes/client-go/tree/v12.0.0/plugin/pkg/client/auth)
2. Include LICENSE file per good practices,

**TODO**:
- Update the installer shell-script to skip the directory path to get to the `kubectl-openebs` binary

improvement: #72, `https://github.com/kubernetes-sigs/krew-index/pull/1460`

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -26,9 +26,10 @@ archives:
       - goos: 'windows'
         format: 'zip'
 
-    wrap_in_directory: true
+    wrap_in_directory: false
     files:
       - README.md
+      - LICENSE
     replacements:
       darwin: Darwin
       linux: Linux
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -26,7 +26,7 @@ archives:
       - goos: 'windows'
         format: 'zip'
 
-    wrap_in_directory: false
+    wrap_in_directory: true
     files:
       - README.md
       - LICENSE
--- a/pkg/client/lvmlocalpv.go
+++ b/pkg/client/lvmlocalpv.go
@@ -26,6 +26,9 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
+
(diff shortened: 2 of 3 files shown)


.goreleaser.yml
```diff
@@ -26,9 +26,10 @@ archives:
       - goos: 'windows'
         format: 'zip'
 
-    wrap_in_directory: true
+    wrap_in_directory: false
```
Abhinandan-Purkait: Let this be true only as we have already submitted the plugin with path.

vharsh: We'd need to re-submit it as likely the krew CI looks for LICENSE file in the root

vharsh: I didn't look for docs which can explicitly specify the relative path of the LICENSE & README.md(these files should also be shipped with the binary).

Abhinandan-Purkait: Re-submit as in? Raise another PR?

vharsh: Yes, but we can keep this true and update the paths for krew-CI spec too, e.g. [plugins/resource-snapshot.yaml](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/resource-snapshot.yaml#L28)

vharsh: > Re-submit as in? Raise another PR?

Yes, we'd have to manually submit once for 0.3.0 if I change the relative path of the binary, later that can be automated via their GitHub action(sooo coool) 😄

Abhinandan-Purkait: Yes. I was looking at their publisher bot. Maybe I will try to set that up. Please do whatever suits best for this PR.

vharsh: I've wrapped the binaries inside the directory, I'll update the configuration for krew by checking with my own krew index.

> Yes. I was looking at their publisher bot.

Yes, we'd do that from 0.4.0 releases.

Abhinandan-Purkait (approved):
LGTM. Thank you.

PR #77, opened 2021-08-13
feat(volume debugger): Add module for CVR events, integration of all modules

### LINKED ISSUE #42: Can the `describe pvc` help to determine - why a cStor Volume is not ready?

Abhinandan-Purkait:
### What does this PR do?
- This PR adds the module for fetching CVR events.
- This integrates all modules and completes the debugger feature.

### Usage and Output
Screenshot from 2021-08-13 10-27-59

### DIFF
--- a/pkg/persistentvolumeclaim/debug_cstor.go
+++ b/pkg/persistentvolumeclaim/debug_cstor.go
@@ -329,7 +329,7 @@ func displayCSPIEvents(k client.K8sClient, crs util.CstorVolumeResources) error 
 			return nil
 		}
 	}
-	return errors.New("no CVRs present to display events")
+	return errors.New("no CSPIs present to display events")
 }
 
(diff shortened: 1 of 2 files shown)


pkg/persistentvolumeclaim/debug_cstor.go
```diff
@@ -328,5 +329,33 @@ func displayCSPIEvents(k client.K8sClient, crs util.CstorVolumeResources) error
 			return nil
 		}
 	}
-	return errors.New("no CSPIs present to display events")
+	return errors.New("no CVRs present to display events")
```
vharsh: This line is likely a part of `displayCSPIEvents`, so likely the previous error message was correct, which will likely run only if crs.CSPIs == nil or has no items.

vharsh (approved):
lgtm

PR #92, opened 2021-08-24
Add CSPC patch to move pool's disks from an older to newer node after disks have moved to newer node

### LINKED ISSUE #84: Add command to suggest the changes that need to be done to CSPC
In cases, where a node is offline or gone out of the cluster and the disks from old (gone) node are moved to a new node in the cluster. Provide the user with an updated CSPC yaml with the required node selector change. 

This can be a follow-up activity or a correction action - once the cstor debug command has identified that there is a problem to a given CSPI.

vharsh:
# UX
```bash
➜  openebsctl git:(cstor_pool_move) ✗ kubectl -n openebs get cspc default-cstor-disk -oyaml | grep old
      kubernetes.io/hostname: old-node

# success case
➜  openebsctl git:(cstor_pool_move) ✗ kubectl openebs update storage default-cstor-disk --from-node=old-node --to-node=minikube
➜  openebsctl git:(cstor_pool_move) ✗ echo $?                                                         
0
➜  openebsctl git:(cstor_pool_move) ✗ kubectl -n openebs get cspc default-cstor-disk -oyaml | grep old
# the old-node has been changed to `minikube` node which is in Ready state after the successful patch operation

# failure case
➜  openebsctl git:(cstor_pool_move) ✗ kubectl openebs update storage default-cstor-disk --from-node=minikube --to-node=dummynode 
node dummynode not found
```

# Questions

- Is the current command structure okay for this or should a `node` or a `pool-migrate` sub-command/word be inserted somewhere?
- Is it okay to put in some *more* small nifty tasks which **modify the state of the system** which the end user had to do it anyway(maybe with a more steady hand and after some thought & checks)?
- Is this implementation okay, considering the issue(#84) description clearly states that the CLI should provide the updated YAML? `Provide the user with an updated CSPC yaml with the required node selector change.`

fixes: #84
Co-authored-by: Shubham Bajpai (@shubham14bajpai) <shubham.bajpai@mayadata.io>

### DIFF
--- a/pkg/storage/cstor_update.go
+++ b/pkg/storage/cstor_update.go
@@ -84,7 +84,7 @@ func DebugCSPCNode(k *client.K8sClient, cspc string) (map[string]string, error) 
 		}
 	}
 	// 3. Fetch all the BDs
-	bds, err := k.GetBDs(devices, "")
+	bds, _ := k.GetBDs(devices, "")
 	actualBDToHost := make(map[string]string)
 	for _, bd := range bds.Items {
 		actualBDToHost[bd.Name] = bd.Labels["kubernetes.io/hostname"]
--- a/pkg/storage/cstor_update.go
+++ b/pkg/storage/cstor_update.go
@@ -14,9 +14,6 @@ import (
 
 // CSPCnodeChange helps patch the CSPC for older nodes
 func CSPCnodeChange(k *client.K8sClient, poolName, oldNode, newNode string) error {
-	if k == nil {
-		k, _ = client.NewK8sClient("openebs")
-	}
 	cspc, err := k.GetCSPC(poolName)
 	if err != nil {
 		return fmt.Errorf("CStor pool cluster %s not found", poolName)
--- a/pkg/storage/cstor_update.go
+++ b/pkg/storage/cstor_update.go
@@ -1,0 +1,45 @@
+package storage
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	cstorv1 "github.com/openebs/api/v2/pkg/apis/cstor/v1"
+	"github.com/openebs/openebsctl/pkg/client"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+)
+
+// CSPCnodeChange helps patch the CSPC for older nodes
+func CSPCnodeChange(k *client.K8sClient, poolName, oldNode, newNode string) error {
+	if k == nil {
+		k, _ = client.NewK8sClient("openebs")
+	}
+	cspc, err := k.GetCSPC(poolName)
+	if err != nil {
+		return fmt.Errorf("CStor pool cluster %s not found", poolName)
+	}
+	var newPool cstorv1.CStorPoolCluster
+	_, err = k.GetNode(newNode)
+	if err != nil {
+		return fmt.Errorf("node %s not found", newNode)
+	}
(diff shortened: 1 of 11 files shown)


pkg/storage/storage.go
```diff
@@ -126,3 +126,17 @@ func CasDescribeMap() map[string]func(*client.K8sClient, string) error {
 func CasDescribeList() []func(*client.K8sClient, string) error {
 	return []func(*client.K8sClient, string) error{DescribeCstorPool, DescribeZFSNode, DescribeLVMvg}
 }
+
+// Update is the entrypoint to manage the Update of storage resources of a particular cas-type
+func Update(namespace, storageName, initial, final string) error {
```
vharsh: To add support for more cas-types, one can add the optional casType argument and ensure that update operation for the resource of that cas-type is called, this might help with some smaller instances(preferably non critical) of common day-1 or day-2 operations.

pkg/storage/storage.go
```diff
@@ -126,3 +126,17 @@ func CasDescribeMap() map[string]func(*client.K8sClient, string) error {
...
+	// 1. Currently only Cstor's node patch is supported
+	k, _ := client.NewK8sClient(namespace)
+	if namespace == "" {
+		ns, err := k.GetOpenEBSNamespace(util.CstorCasType)
+		k.Ns = ns
+	if err != nil {
+			return fmt.Errorf("unable to detect cstor, please specify cstor's --openebs-namespace")
```
vharsh, line 152: observed indentation was a bit off, fixed it via vim-go, no wonder CI missed it 🤔

pkg/storage/cstor_update.go
```diff
@@ -0,0 +1,47 @@
+package storage
```
vharsh, line 17: Need to check why license check didn't catch this file as the license was missing earlier.

kmova:
@vharsh -- can we provide the UX via the following:

```
kubectl openebs describe storage default-cstor-disk --openebs-namespace=openebs --debug
```

Now, the node that went offline can be obtained from the above. It will be good to see if we can provide the suggested node -- based on the BDs in that pool. The command can be explicit like:

```
kubectl openebs update storage default-cstor-disk --openebs-namespace=openebs --from-node=old-node-name --to-node=new-node-node
```

vharsh:
Sure, I understand the `update` subcommand might mean a lot of things. I'll quickly add from-node & to-node flags and add sufficient error handling in place.

vharsh:
@kmova Yes, this(debug command enhancement) can be done.
There's something inconsistent here in terms of usability, we consider a CSPI as `storage` but while editing, we consider the `storage` as a CSPC. Should we reconsider what's a `storage` for cstor or should we use the poolInstance in this command too(as it's the actual pool by definition).

vharsh:
@kmova I had some minor work done on this earlier and forgotten about it, I've now pushed all the changes, this has been removed from RC1 in yesterday's community call but this PR can be reviewed if time permits.

I haven't tested the debug functionality on a K8s cluster, but I have ~88% code coverage so I'm confident that the code will require minimal changes due to potential hidden bugs. I have missed covering around 3 lines(one branch, thanks to CodeCov❤️), which I'll do it by EoD.

vharsh:
Converting it back to draft and pushing the bug-PoC, this logic has a minor flaw, If a pool has two instances with two disks each and **ONLY one of the BDs** is out of place, it'd go ahead and patch the CSPC, this needs to be avoided as the user should be able to move back the 1 out of 4 out-of-place BD back to the correct node instead of moving the CSPC to get to the same quantitative state where 3 out of 4 BDs are in their correct place.

Abhinandan-Purkait:
Closing this since cstor is to be deprecated now. Thanks.

PR #95, opened 2021-08-30
feat(cluster-info): Add feature to show component versions, statuses and running components for each installed engine

### LINKED ISSUE #36: Add support to enlist component-versions & statuses of OpenEBS
# Component-versions

## Scope

As of now, there's a sub-command `version` which helps end users get info about the version of this `kubectl-plugin`, just like kubectl it can be expanded to give in more information about installed OpenEBS versions, etc from component labels.

## TUI UX

```bash
// need to discuss with the community for better usability & debug-ability
$ kubectl openebs cluster-info.
CAS-TYPE       VERSION        STATUS
CStor            3.0             Good
LocalPV          2.9             Upgrade available
CStor-Legacy     1.4             Good
```

Much later, more functionality can be added which can raise more potential red-flags besides quality information, like `kubectl get cs` or `kubectl cluster-info`, a good source could be the `ERROR/WARNING` events from the Storage provisioners, CSI pods, etc, if it's not super easy to do so via `kubectl get events --<flags>`

Abhinandan-Purkait:
### What does this PR do?
- This PR adds the `cluster-info` command to show the control-plane version and status.
- This adds the template to add any `cas-type` to this command.
- The `WORKING` column shows the number of control-plane components running vs no. of components should be Running.

### Usage and Output
```
$ kubectl openebs cluster-info
```  
Screenshot from 2021-08-30 11-07-38

Closes #36

### DIFF
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -703,3 +703,11 @@ func (k K8sClient) GetEvents(fieldSelector string) (*corev1.EventList, error) {
 	}
 	return events, nil
 }
+
+func (k K8sClient) GetPods(labelSelector string, fieldSelector string, namespace string) (*corev1.PodList, error) {
+	pods, err := k.K8sCS.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labelSelector, FieldSelector: fieldSelector})
+	if err != nil {
+		return nil, fmt.Errorf("error getting pods : %v", err)
+	}
+	return pods, nil
+}
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -696,6 +696,7 @@ func (k K8sClient) GetCSIControllerSTS(name string) (*appsv1.StatefulSet, error)
 	}
 }
 
+// GetEvents returns the corev1 events based on the fieldSelectors
 func (k K8sClient) GetEvents(fieldSelector string) (*corev1.EventList, error) {
 	events, err := k.K8sCS.CoreV1().Events("").List(context.TODO(), metav1.ListOptions{FieldSelector: fieldSelector})
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/client/k8s.go
```diff
@@ -703,3 +703,11 @@ func (k K8sClient) GetEvents(fieldSelector string) (*corev1.EventList, error) {
 	}
 	return events, nil
 }
+
```
vharsh, line 707: Please add a comment for this exported function, I wonder how golint is passing 🤔

pkg/cluster-info/cluster-info.go
```diff
@@ -0,0 +1,140 @@
...
+			version := getVersion(componentDataMap)
+			namespace := getNamespace(componentDataMap)
+			clusterInfoRows = append(
+				clusterInfoRows,
+				metav1.TableRow{Cells: []interface{}{casType, namespace, version, working, util.ColorStringOnStatus(status)}},
+			)
+		}
+	}
+	if len(clusterInfoRows) == 0 {
+		fmt.Println("None Of the OpenEBS Storage Engines are installed in this cluster")
```
vharsh: Also return an error to the caller so that error codes can be set accordingly by `cmd.Run`, after this change, the else will not be required(gosimpl will throw errors)

pkg/cluster-info/cluster-info.go
```diff
@@ -0,0 +1,140 @@
...
+						CasType:   casType,
+					}
+				}
+			} else {
+				componentDataMap[item.Labels["openebs.io/component-name"]] = util.ComponentData{
+					Namespace: item.Namespace,
+					Status:    string(item.Status.Phase),
+					Version:   item.Labels["openebs.io/version"],
+					CasType:   casType,
+				}
```
vharsh, line 86: These appear equivalent, maybe nesting functionality inside an if-else isn't required.

vharsh (approved):
lgtm, minor suggestions

pkg/cluster-info/cluster-info.go
vharsh: I mean the functionality in `if` & `else` are same

vharsh: Ah silly me, I missed that the inner if is not associated with the else which is associated with the outer if.

vharsh (approved):
lgtm

PR #96, opened 2021-08-30
feat: add support for pushing test-coverage metrics via codecov

### LINKED ISSUE #61: Add support for running tests & pushing test-coverage metrics
Code Coverage is a metric to see how much of the code by means of entities viz statements, branches, functions, etc run when the tests are run. It is a good metric to see how much of the business logic is tested(the expected & actual values are checked assuming no other measurable side effect) & it helps ensure that an existing functionality does not break when newer ones are added or older ones are refactored. While it's not the absolute metric of code quality, a higher code coverage usually implies well written code with minimal side-effects.

Currently tests are to be written down(tracked @ #49), so it'd be great
- [x] to run them on every Pull-Request & PR merge(I believe that'd be push/comm
- [x] to have **informational** alerts about them going UP/DOWN when PRs are raised, so that people feel encouraged to write tests(automatically managed by CodeCov), as we are ramping up unit-testing, it might not be a very good idea for the CI to fail because code coverage is less than some **X%**
- [x] to have badges on README.md, badges are cool 🎉

Good starting research points:
- CodeCov is a go-to tool for many projects, but are there any alternatives which might be better? 
- Is it better to use the CodeCodv bash exporter or the CodeCov github action, if codecov is the choice of tool?
- Likewise, should running UTs be a workflow separate from the workflow which runs code Linters(GolangCI-Lint)?

Good starting points:
- [GitHub actions docs](https://docs.github.com/en/actions/quickstart)
- [OpenEBS NDM example to run tests & upload to CodeCov](https://github.com/openebs/node-disk-manager/blob/master/.github/workflows/build.yml#L48)

Great readings:
- [How does the code coverage even work in Golang](https://blog.golang.org/cover)
- [How much code-coverage?](https://stackoverflow.com/questions/90002/what-is-a-reasonable-code-coverage-for-unit-tests-and-why)

ParthS007:
closes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/mum4k/termdash/blob/master/LICENSE)
 [![Releases](https://img.shields.io/github/downloads/openebs/openebsctl/total.svg)](https://github.com//openebs/openebsctl/releases)
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fopenebs%2Fopenebsctl.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fopenebs%2Fopenebsctl?ref=badge_shield)
-
+[![codecov.io](https://codecov.io/github/openebs/openebsctl/coverage.svg?branch=develop)](https://codecov.io/github/openebs/openebsctl?branch=develop)
 
 
 OpenEBSCTL is a kubectl plugin to manage OpenEBS storage components.
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -1,0 +1,22 @@
+name: tests and coverage
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - develop
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+      - name: Run coverage
+        run: go test -coverprofile=coverage.out -covermode=atomic
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2


.github/workflows/golangci-test.yml
```diff
@@ -0,0 +1,24 @@
+name: golangci-test
```
vharsh: `tests and coverage` might be a better name for this workflow, golangci-lint is the name of the project which runs other linters, also consider changing the filename to something similar

.github/workflows/golangci-test.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.14'
```
vharsh: Please bump this to `1.16` which is being used in other action workflows, sometime later, we'd need to change it to `1.17` as in all workflows & go.mod file.
```suggestion
          go-version: '1.16'
```

.github/workflows/golangci-test.yml
```diff
@@ -0,0 +1,24 @@
...
+      - name: Run coverage
+        run: go test -coverprofile=coverage.out -covermode=atomic
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
+        files: ./coverage.out
```
vharsh: There are some errors at [this line](https://github.com/openebs/openebsctl/actions/runs/1181384766/workflow) PTAL, I don't think a `CODECOV_TOKEN` is required to be set as this is a public repository.

I think, it should be `file` inside the `with` object/key, like [this](https://github.com/codecov/codecov-action/blob/master/.github/workflows/main.yml#L22)

.github/workflows/golangci-test.yml
```diff
@@ -0,0 +1,24 @@
...
+        flags: unittests
```
vharsh: Awesome #TIL for me 👍🏼

README.md
```diff
@@ -9,7 +9,7 @@
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://github.com/mum4k/termdash/blob/master/LICENSE)
 [![Releases](https://img.shields.io/github/downloads/openebs/openebsctl/total.svg)](https://github.com//openebs/openebsctl/releases)
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Fopenebs%2Fopenebsctl.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Fopenebs%2Fopenebsctl?ref=badge_shield)
-
+[![codecov.io](https://codecov.io/github/openebs/openebsctl/coverage.svg?branch=develop)](https://codecov.io/github/openebs/openebsctl?branch=develop)
```
vharsh, line 12: 👍🏼

vharsh (commented):
thanks for your research & implementation 🎉 😃

.github/workflows/golangci-test.yml
Abhinandan-Purkait: How is the codecov token configured?

vharsh: Per the docs, CODECOV_TOKEN [isn't required](https://github.com/codecov/codecov-action#usage) for public repositories if the CI is GitHub actions, CircleCI, etc.

vharsh (approved):
lgtm

PR #97, opened 2021-08-30
fix(codecov): Correct the command to run go test

### LINKED ISSUE #61: Add support for running tests & pushing test-coverage metrics

Abhinandan-Purkait:

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -17,6 +17,6 @@ jobs:
         with:
           go-version: '1.16'
       - name: Run coverage
-        run: go test -coverprofile=coverage.out -covermode=atomic
+        run: go test ./... --coverprofile=coverage.out --covermode=atomic
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v2


vharsh (approved):
lgtm

PR #99, opened 2021-09-02
enhacement: Notify with latest version of openebsctl

Abhishek-kumar09:
$ kubectl openebs version will now be able to notify if the used version is expired.
screenshot:
Screenshot from 2021-09-02 17-51-38


cmd/version/version.go
```diff
@@ -78,8 +84,64 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
 	}
 	cmd.SetUsageTemplate(versionCmdHelp)
 	return cmd
 }
+
+func getCliLatestVersion(currVersion string) {
+	// getting the latest version of openebsctl from sigs.k8s.io/krew-index
+	resp, err := http.Get("https://raw.githubusercontent.com/kubernetes-sigs/krew-index/master/plugins/openebs.yaml")
+	util.CheckErrDefault(err, "Error reading response body")
+	defer resp.Body.Close()
```
vharsh, line 101: this should only be done if `err == nil`, i.e. before the error gets checked

cmd/version/version.go
```diff
@@ -78,8 +84,64 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
+	body, err := ioutil.ReadAll(resp.Body)
+	util.CheckErrDefault(err, "Error reading response body")
+
+	var data map[string]interface{}
+	err = yaml.Unmarshal(body, &data)
+	util.CheckErrDefault(err, "Error parsing yaml")
+
+	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
+	if !isLatestVersion(currVersion, latestVersion) {
+		fmt.Printf("\nYou are using older version (%s) of cli, latest available version is %s\n", currVersion, latestVersion)
```
vharsh: Use `fmt.Println`, this CLI project can also run on non Nix shells which prefer carriage returns due to historical and other reasons.

cmd/version/version.go
```diff
@@ -78,8 +84,64 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
 		},
 	}
 	cmd.SetUsageTemplate(versionCmdHelp)
 	return cmd
 }
+
+func getCliLatestVersion(currVersion string) {
+	// getting the latest version of openebsctl from sigs.k8s.io/krew-index
+	resp, err := http.Get("https://raw.githubusercontent.com/kubernetes-sigs/krew-index/master/plugins/openebs.yaml")
+	util.CheckErrDefault(err, "Error reading response body")
```
vharsh:
```suggestion
      if err != nil {
         return something
    }
```
you might not need to pass the error around for checking the error, 

Don't commit it from here, do it from local machine

cmd/version/version.go
```diff
@@ -78,8 +84,64 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
 
 			util.TablePrinter(util.VersionColumnDefinition, rows, printers.PrintOptions{Wide: true})
+			getCliLatestVersion(rootCmd.Version)
 		},
 	}
 	cmd.SetUsageTemplate(versionCmdHelp)
 	return cmd
 }
+
+func getCliLatestVersion(currVersion string) {
```
vharsh: I'm sorry for being a grammar nazi here, but the name `getCliLatestVersion` makes me think this function should return the latest version, maybe checking for the latest & current function for differences can happen in another function or maybe this can be renamed as `checkForLatestVersion`.

cmd/version/version.go
```diff
@@ -78,8 +84,64 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
+	err = yaml.Unmarshal(body, &data)
+	util.CheckErrDefault(err, "Error parsing yaml")
+
+	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
+	if !isLatestVersion(currVersion, latestVersion) {
+		fmt.Printf("\nYou are using older version (%s) of cli, latest available version is %s\n", currVersion, latestVersion)
+	}
+}
+
+func isLatestVersion(currVersion string, latestVersion string) bool {
```
vharsh, line 128: How about just returning `currVersion == latestVersion`, it'd work just about right as latestVersion would be updated via a GitHub action from now on

pkg/util/error.go
```diff
@@ -40,3 +40,16 @@ func CheckErr(err error, handleErr func(string)) {
 	}
 	handleErr(err.Error())
 }
+
+// CheckErrDefault to handle command errors with default handler
+func CheckErrDefault(err error, msg string) {
```
vharsh: This function is not super required IMO, for golang, devs usually prefer to handle the error in the same place where they are generated, this might seem a bit counter-intuitive at first.

Ref: https://go.dev/blog/error-handling-and-go
Ref: https://go.dev/blog/errors-are-values

vharsh (commented):
awesome feature, got some minor comments

cmd/version/version.go
vharsh: so if for some reason, `err` is non-nil maybe the internet connection is down, etc the `resp` will be `nil` and the code will panic, it'd look something like a NullPointerException

cmd/version/version.go
vharsh: iirc, macs use `\r\n` for the next line & windows use `\r` & linux uses `\n`, it's best to let libraries manage this chaos 😹

cmd/version/version.go
Abhishek-kumar09: Yes, corrected.

cmd/version/version.go
Abhishek-kumar09: Oh, Yeah done. 
I am using Println now, but used `\n` in the Println to get some padding from the previous command output. Is that fine?

cmd/version/version.go
Abhishek-kumar09: Okay, deleted it :smile:

cmd/version/version.go
Abhishek-kumar09: Yeah right, my mistake xD. Changed.

cmd/version/version.go
Abhishek-kumar09: actually right, this is great.

pkg/util/error.go
Abhishek-kumar09: Removed, Thanks.

cmd/version/version.go
vharsh: No I mean, something like [this](https://golangbyexample.com/resposne-body-closed-golang/) and remove the `CheckErr` function, most go-code has error handling code in the same place.

Besides, if CheckErr doesn't terminate the program there'd be a panic at line 96 because `resp` will be nil.

vharsh: i.e. `resp.Body.Close()` should be only called if err is nil, i.e. resp is non-nil, usually if err is non-nil the structure/value returned is a zero-value,

cmd/version/version.go
```diff
@@ -78,8 +82,48 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
 }
+
+func checkForLatestVersion(currVersion string) {
+	// getting the latest version of openebsctl from sigs.k8s.io/krew-index
+	resp, err := http.Get("https://raw.githubusercontent.com/kubernetes-sigs/krew-index/master/plugins/openebs.yaml")
+	util.CheckErr(err, util.Fatal)
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	util.CheckErr(err, util.Fatal)
```
vharsh: return err if err is not nil, though this will likely not be the case unless `resp.Code != 200` i.e. the response to this request is `HTTP 200 OK`

cmd/version/version.go
Abhishek-kumar09: Got it, Will do.
Actually the current version of code also works fine, since we were calling `util.Fatal` on error, calling `util.Fatal` close the program and return thus not calling the defer after that. The program completely exits since fatala has `os.Exit()` call at the end. 
I had checked the code with no internet connection too.

cmd/version/version.go
```diff
@@ -78,8 +82,65 @@ func NewCmdVersion(rootCmd *cobra.Command) *cobra.Command {
...
+
+	if currVersion[0] < '0' || currVersion[0] > '9' {
+		return true
+	}
+
+	if currVersion != latestVersion {
+		return false
+	}
+
+	return true
```
vharsh:
```suggestion
	return latestVersion == currVersion
```

Go has a niche tool called gosimple or something similar name(I cannot recall it right away), this would likely suggest something similar. Lower LoC gets us lesser debugging time :)

Abhishek-kumar09: Actually that was the defensive code:
```
       if currVersion[0] == 'v' {
		currVersion = currVersion[1:]
	}
	if latestVersion[0] == 'v' {
		latestVersion = latestVersion[1:]
	}
```

to check if any of the two versions coming up with v appended in front then they could be checked properly.
But since We are the setter of the version, Are we pretty sure that we will be naming both versions (i.e. current one and latest one with v in front like v0.3.1)?
@vharsh

vharsh: Yes, the currentVersion can have arbitrary things like `dev` or a short commit-hash, but the latestVersion will ideally always begin with `v`.

Note: Updating the plugin version been automated by #87, it won't work the first time without human checks this month(15th September), but later on, it can be expected to work always.

Abhishek-kumar09: Done

vharsh (approved):
lgtm

PR #103, opened 2021-09-05
usability: Handle Exceptions for no resources

### LINKED ISSUE #56: usability: message for no resources found
When the cluster doesn't have required resources, display a helpful message instead of displaying nothing. 

In my cluster that doesn't have any pools setup at, I get the following:
```
kiran_mova_mayadata_io@kmova-dev:~$ kubectl openebs get storage
kiran_mova_mayadata_io@kmova-dev:~$ 
```

It would nice to display a message like this:
```
kiran_mova_mayadata_io@kmova-dev:~$ kubectl get cspc
No resources found in default namespace.
```

Also, while at this - check the message displayed when there is no openebs installed in the cluster.

Abhishek-kumar09:

Fixes #56 

Now empty resource lists are handled by the openebsctl for storage, volumes and bd. They are handled separately on the basis of resourceType, casType and Namespaces.

Screenshot from 2021-09-05 22-10-34

### DIFF
--- a/pkg/blockdevice/blockdevice.go
+++ b/pkg/blockdevice/blockdevice.go
@@ -89,9 +89,8 @@ func createTreeByNode(k *client.K8sClient, bdNames []string) error {
 	}
 	if len(rows) == 0 {
 		return util.HandleEmptyTableError("Block Device", k.Ns, "")
-	} else {
-		// Show the output using cli-runtime
-		util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
 	}
+	// Show the output using cli-runtime
+	util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
 	return nil
 }
--- a/pkg/blockdevice/blockdevice.go
+++ b/pkg/blockdevice/blockdevice.go
@@ -88,7 +88,7 @@ func createTreeByNode(k *client.K8sClient, bdNames []string) error {
 		rows = append(rows, metav1.TableRow{Cells: []interface{}{"", "", "", "", "", "", ""}})
 	}
 	if len(rows) == 0 {
-		util.HandleEmptyTableError("Block Device", k.Ns, "")
+		return util.HandleEmptyTableError("Block Device", k.Ns, "")
 	} else {
 		// Show the output using cli-runtime
 		util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
--- a/pkg/blockdevice/blockdevice.go
+++ b/pkg/blockdevice/blockdevice.go
@@ -87,7 +87,11 @@ func createTreeByNode(k *client.K8sClient, bdNames []string) error {
 		// Add an empty row so that the tree looks neat
 		rows = append(rows, metav1.TableRow{Cells: []interface{}{"", "", "", "", "", "", ""}})
 	}
-	// Show the output using cli-runtime
-	util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
+	if len(rows) == 0 {
+		util.HandleEmptyTableError("Block Device", k.Ns, "")
+	} else {
+		// Show the output using cli-runtime
+		util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
+	}
 	return nil
 }
--- a/pkg/util/error.go
+++ b/pkg/util/error.go
@@ -40,3 +40,16 @@ func CheckErr(err error, handleErr func(string)) {
 	}
 	handleErr(err.Error())
 }
+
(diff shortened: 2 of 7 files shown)


pkg/blockdevice/blockdevice.go
```diff
@@ -87,7 +87,11 @@ func createTreeByNode(k *client.K8sClient, bdNames []string) error {
...
 		rows = append(rows, metav1.TableRow{Cells: []interface{}{"", "", "", "", "", "", ""}})
 	}
-	// Show the output using cli-runtime
-	util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
+	if len(rows) == 0 {
+		return util.HandleEmptyTableError("Block Device", k.Ns, "")
+	} else {
+		// Show the output using cli-runtime
+		util.TablePrinter(util.BDTreeListColumnDefinations, rows, printers.PrintOptions{Wide: true})
+	}
```
vharsh: Dang, I'm positive golangCI-lint isn't running all the tools now, I'll will create a `good-first-issue` for the same.

# What should have happened if CI was working optimally ?

The build should have ideally failed because we have added GolangCI-lint to our CI, I cannot recall the exact warning text but the functionality won't change even if the else bracket is removed, the tool which does it likely gosimple.

```go
func awesome() {
    ....
    if something {
        return nil, err
    } else {
        return value, nil
    }
} // end of function
```

can get simplified to

```go
func awesome() {
    ....
    if something {
        return nil, err
    }
    return value
} // end of function
```

pkg/util/error.go
```diff
@@ -40,3 +40,16 @@ func CheckErr(err error, handleErr func(string)) {
 	}
 	handleErr(err.Error())
 }
+
+// Handle Empty Table Error
```
vharsh:
```suggestion
// HandleEmptyTableError handles errors when a resource or set of resources aren't found
```

This is a minor linting issue in golang, exported functions must be properly documented as such.
Ref: https://go.dev/blog/godoc

pkg/util/error.go
```diff
@@ -40,3 +40,16 @@ func CheckErr(err error, handleErr func(string)) {
...
+func HandleEmptyTableError(resource string, ns string, casType string) error {
+	if ns == "" && casType == "" {
+		return fmt.Errorf("no %s found in your cluster", resource)
+	} else if ns != "" && casType != "" {
+		return fmt.Errorf("no %s %s found in %s namespace", casType, resource, ns)
+	} else if casType != "" {
```
vharsh: Abhinandan has recently written a nice utility function which returns true for valid casTypes. Try to rebase your code against the current `origin/develop` and you'd likely see that in this package. Use that one for this condition check.

pkg/util/error_test.go
```diff
@@ -57,3 +57,53 @@ func TestCheckErr(t *testing.T) {
...
+			args{
+				resource: "ResourceType",
+				ns:       "InValidNamespace",
+				casType:  "",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_ = HandleEmptyTableError(tt.args.resource, tt.args.ns, tt.args.casType)
```
vharsh: This is an awesome first test and it likely covers all branches too, but can you compare the error values, i.e. the ones which are returned? That'd be just awesome, for that check the error interface once, you'd find an easy way to compare the expected error from the actual error.

Currently this test doesn't check the return values from the function, i.e. if the code coverage has increased but the real value of unit testing hasn't been put to use to detect if the `system-under-test`(fancy name for the code being tested) 's functionality has changed by a bit and it's callers must be wary of how it's change can affect their functionality.

vharsh (commented):
Awesome first unit test. 🎉 😺
This is a cool use of a function for writing out errors predictably. PTAL some minor suggestions for the UT.

pkg/blockdevice/blockdevice.go
Abhishek-kumar09: Corrected.

pkg/util/error.go
Abhishek-kumar09: Sounds great, Done

pkg/util/error.go
Abhishek-kumar09: Done

pkg/util/error_test.go
Abhishek-kumar09: Yes, thanks for the help. Just modified it please have a look.

pkg/util/error.go
```diff
@@ -40,3 +40,16 @@ func CheckErr(err error, handleErr func(string)) {
 	}
 	handleErr(err.Error())
 }
+
+// Handle Empty handles error when resources or set of resources are not found
```
vharsh:
```suggestion
// HandleEmptyTableError handles error when resources or set of resources are not found
```

vharsh (approved):
lgtm

PR #107, opened 2021-09-07
bug(describe): Fix bug to show error if describe is invoked without a name

Abhinandan-Purkait:


pkg/persistentvolumeclaim/debug.go
```diff
@@ -27,7 +27,7 @@ import (
 
 // Debug manages various implementations of PersistentVolumeClaim Describing
 func Debug(pvcs []string, namespace string, openebsNs string) error {
-	if pvcs == nil {
+	if len(pvcs) == 0 && pvcs == nil {
```
vharsh:
```suggestion
	if len(pvcs) == 0 || pvcs == nil {
```

vharsh (approved):

PR #109, opened 2021-09-07
golangci-lint is not working in vscode

### LINKED ISSUE #105: Investigation: Not all tools of GolangCI lint are running in the CI
There is likely some mis-configuration in the CI which is ignoring minor linters like gosimpl, golint, etc from raising red flags. E.g.

```go
func awesome() {
    ....
    if something {
        return nil, err
    } else {
        return value, nil
    }
} // end of function
```
should raise warnings to the very least, but it isn't being raised currently.

# Starting points for completion

1. Take the [Go-Tour](https://tour.golang.org) if you aren't up to speed with Golang
2. Read some basics about Continuous Integration and something about [GolangCI Lint](https://golangci-lint.run) and what [linters](https://golangci-lint.run/usage/linters) it supports.
3. Enable some of the linters like `revive`, `gofmt`, `errorlint`, etc.
4. Fix the linting issues

Amishakumari544:
Tried to check If this enable is working fine or not but It's not working in VSCode

Screenshot (8)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,3 +18,16 @@ linters-settings:
       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
       See the License for the specific language governing permissions and
       limitations under the License.
+
+enable:
+    - bodyclose
+    - deadcode
+    - errcheck
+    - goconst
+    - gocritic
+    - gocyclo
+    - gofmt
+    - goimports
+    - golint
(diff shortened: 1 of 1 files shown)


vharsh (approved):
lgtm, this might be passing due to some action caching issue

Having these linters run, will likely trigger errors in upcoming code changes. Thanks for the fix 🎉 😄

PR #110, opened 2021-09-09
feat(localhostpath) Add support for local-hostpath volume list and describe

### LINKED ISSUE #106: Add support for listing and describing hostpath volumes
### Current Behaviour
```
 ❯ kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                                                                 STORAGECLASS          REASON   AGE
pvc-69e44403-1b3c-4028-9dd8-d3db547240d6   4Gi        RWO            Delete           Bound    openebs/openebs-pvc-7f194739-16c8-4920-bb79-a28d63183b3e-jiva-rep-0   openebs-hostpath               47h
pvc-7f194739-16c8-4920-bb79-a28d63183b3e   4Gi        RWO            Delete           Bound    default/example-jiva-csi-pvc                                          openebs-jiva-csi-sc            47h

```
```
 ❯ ./k describe volume pvc-69e44403-1b3c-4028-9dd8-d3db547240d6
could not determine the underlying storage engine ns, please provide using '--openebs-namespace' flag

```

### Expected Behaviour
It should describe the hostpath volume

Abhishek-kumar09:
Screenshot from 2021-09-09 23-25-02

 An initial implementation to describing and list local-hostpath volumes


pkg/util/constant.go
```diff
@@ -37,6 +37,8 @@ const (
 	LocalHostpathCasType = "localpv-hostpath"
 	// LocalDeviceCasType cas type name
 	LocalDeviceCasType = "localpv-device"
+	// Local host path cas type name
+	LocalHostpath = "local-hostpath"
```
Abhishek-kumar09: `LocalHostpath = "local-hostpath"` and `LocalHostpathCasType = "localpv-hostpath"`
are bit confusing in the first look and maybe error prone in future, Can We change the name, because `LocalHostpath` actually shows the cas type of Local-Hostpath

kmova: question. why are we adding a new type called "LocalHostpath = "local-hostpath"?

Abhishek-kumar09: Because that is the cas-type for showing local-hostpath volumes.
When Jiva volumes are created, they create the underlying replicas and PV the PV that it creates is the `local-hostpath` with the storage class name as `openebs-hostpath`.  This volume is actual Persistent volume that has a claim of the jiva volumes. (acc to my understanding)
Earlier We could not describe the `localhostpath` cas types, this PR does the work to list and describe those PVs.

kmova: Hmm.. so is the following accurate:
- localpv-hostpath ( will display the details for PVCs created by users)
- local-hostpath ( will be used for displaying backend PVCs?) 

Can we extend `localpv-hostpath` itself to cover both cases. Can we do this, if we set a annotation via the jiva storage classes that says, that PVC is a backend PVC for a given jiva PVC?

Let us see if we can avoid introducing additional cas type in this cas.

Abhishek-kumar09: What I think is:
- local-hostpath is the actual local volume that is provisioned by openebs
- It's case type is: `local-hostpath`
- It has path on the cluster's disk where the data resides.
- So it the the PV and other CSI have a claim over this volume

Currently having less knowledge about openebs-localPV, but combining make more sense if both of them are doing the same thing. I think @vharsh could clarify further details with more clarity :smiley:

pkg/util/types.go
```diff
@@ -86,6 +86,13 @@ type VolumeInfo struct {
 	JVP string
 }
 
+type LocalHostPathVolInfo struct {
+	VolumeInfo
+	Path string
+	ReclaimPolicy string
+	CasType string
+}
```
vharsh: Please run `gofmt -s -w pkg/util/types.go`, either this tool is not correctly installed on your system or your IDE isn't able to find it.

If you are using vscode, please follow steps from [this StackOverflow answer](https://stackoverflow.com/a/53513867)

pkg/volume/volume.go
```diff
@@ -103,7 +103,7 @@ func Describe(vols []string, openebsNs string) error {
 		if openebsNs == "" {
 			if val, ok := nsMap[casType]; ok {
 				k.Ns = val
-			} else if casType != util.ZFSCasType && casType != util.LVMCasType {
+			} else if casType != util.ZFSCasType && casType != util.LVMCasType && casType != util.LocalHostpath {
```
vharsh: I had almost forgot why I had this code-quirk, serves me right for incomplete code-comments🤣 

IIRC, this is required because the CSI provisioners and _now the external-provisioners_ of these volumes are present in `kube-system` namespace and thus if `GetOpenEBSNamespaceMap()` had to figure out `cas-type→namespace relation`, it'd likely figure it out incorrectly as the namespace will not be `kube-system` i.e. the relevant custom resources if any would be present on some other namespace where the provisioner is actually present.


As of now, my memory is a bit hazy and the error message and the core around it seems a bit off, I'll re-read and confirm if this code change is required.

pkg/volume/volume.go
```diff
@@ -122,27 +122,29 @@ func Describe(vols []string, openebsNs string) error {
...
 
 // CasDescribeMap returns a map cas-types to functions for volume describing
 func CasDescribeMap() map[string]func(*client.K8sClient, *corev1.PersistentVolume) error {
 	// a good hack to implement immutable maps in Golang & also write tests for it
 	return map[string]func(*client.K8sClient, *corev1.PersistentVolume) error{
-		util.JivaCasType:  DescribeJivaVolume,
-		util.CstorCasType: DescribeCstorVolume,
-		util.ZFSCasType:   DescribeZFSLocalPVs,
-		util.LVMCasType:   DescribeLVMLocalPVs,
+		util.JivaCasType:   DescribeJivaVolume,
```
vharsh: This code is not gofmted

pkg/volume/local_hostpath.go
```diff
@@ -0,0 +1,89 @@
...
+		// Ignore all the other volumes that is not of cas-type local-hostpath
+		if pv.Labels["openebs.io/cas-type"] != util.LocalHostpath {
+			continue
+		}
+
+		name := pv.Name
+		capacity := pv.Spec.Capacity.Storage()
+		sc := pv.Spec.StorageClassName
+		attached := pv.Status.Phase
+		attachedNode, customStatus, ns, storageVersion := pv.Labels["nodeID"], "N/A", "N/A", "N/A"
```
vharsh, line 56:
```suggestion
              customNs := // Get deployments from all namespace which have Label="openebs.io/component-name=openebs-localpv-provisioner"
              customStatus = ""
              version = provisionerDeployment.Label["openebs.io/version"]
		attachedNode, customStatus, ns, storageVersion := pv.Labels["nodeID"], "", "N/A", "N/A"
```

@kmova What's the best way to express the status of LocalPV volumes(plain localpv ones), should it be `"N/A"` as it's technically not-applicable or is a simple `""` slightly better on the eyes?

vharsh: The storage version for a LocalPV hostpath and a LocalPV device can be considered to be the storage version of the provisioner, we've gone forward with this assumption for ZFS-LocalPV & LVM-LocalPV too.

pkg/util/constant.go
vharsh: IMO, another cas-type `local-hostpath` is not required and might as well be confusing. Jiva volume provisioner creates a replicated volume backed by `n` local volumes on different nodes which are provisioned by [dynamic-localpv-provisioner](https://github.com/openebs/dynamic-localpv-provisioner/).

If a user provisions a Local hostpath backed volume, they are provisioned by the same [dynamic-localpv-provisioner](https://github.com/openebs/dynamic-localpv-provisioner/) just that now it's happening directly(not via jiva-operator component).

vharsh: @kmova I need to analyze once why `localpv-hostpath` was present earlier. We wouldn't need both `local-hostpath` and `localpv-hostpath` as both of them are one-and-the-same-thing.



But as an enhancement, we can add some flag or key in the describe output of LocalPV-hostpath which can provide some clarity towards the case you've pointed out, so that the user can ascertain that the first one is something some non-OpenEBS application(like MySQL, MongoDB consuming localpv-hostpath volumes) is directly using and the second one is something created by OpenEBS provisioners to support some non-OpenEBS application(like MySQL, MongoDB consuming Jiva volumes). 

> 
    - localpv-hostpath ( will display the details for PVCs created by users)
    - local-hostpath ( will be used for displaying backend PVCs?)

pkg/util/types.go
Abhishek-kumar09: Okay, I will make sure I run go fmt before every push. Ctrl+s is working but I had autoSave enabled in vscode, so rarely have to explicitly write ctrl+save

Abhishek-kumar09:
newss
With blank fields:
ssnew

pkg/volume/volume.go
Abhishek-kumar09: This was formated by the golang, since there was new addition with slightly bigger name hence, these are shifted a bit ahead.

pkg/util/constant.go
kmova: We can enhance the jiva operator code to put an annotation of something on the PVC/PV launched for jiva replicas. Can you please check if it doesn't exist already.

pkg/volume/local_hostpath.go
```diff
@@ -0,0 +1,94 @@
...
+RECLAIM POLICY  : {{.ReclaimPolicy}}
+`
+)
+
+// GetLocalHostpath returns a list of local-hostpath columes
+func GetLocalHostpath(c *client.K8sClient, pvList *corev1.PersistentVolumeList, openebsNS string) ([]metav1.TableRow, error) {
+	var rows []metav1.TableRow
+	for _, pv := range pvList.Items {
+		// Ignore all the other volumes that is not of cas-type local-hostpath
+		if pv.Labels["openebs.io/cas-type"] != util.LocalHostpath {
```
Abhinandan-Purkait: Use, pkg/utils/k8s_utils.go's `GetCasTypeFromPV` here, as there can be several variations.

pkg/client/k8s.go
```diff
@@ -743,3 +743,21 @@ func (k K8sClient) GetPods(labelSelector string, fieldSelector string, namespace
...
 	return pods, nil
 }
+
+/*
+	LOCAL VOLUMES SPECIFIC METHODS
+*/
+
+// GetLocalPvDeployment returns the deployments with a specific
+// openebs-component-name label key
+func (k K8sClient) GetLocalPvDeployment(name string) (*appsv1.Deployment, error) {
```
Abhinandan-Purkait: This, function can be made generic by, taking the labelSelector as input. Please make sure the code you write can be reused later.

pkg/client/k8s.go
```diff
@@ -743,3 +743,21 @@ func (k K8sClient) GetPods(labelSelector string, fieldSelector string, namespace
...
+	if pv, err := k.K8sCS.AppsV1().Deployments("").List(context.TODO(), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("openebs.io/component-name=%s", name),
+	}); err == nil && len(pv.Items) == 1 {
+		return &pv.Items[0], nil
```
vharsh: @Abhinandan-Purkait Should the function just return `*appsv1.DeploymentList, error` instead, I think we did similar kind of enhancements.

Abhinandan-Purkait: Yes, it can. At the point of calling we can extract the first element, as per this use case.

vharsh:
@Abhishek-kumar09 One test is failed by design and can be fixed easily just change [this](https://github.com/openebs/openebsctl/blob/develop/pkg/volume/volume_test.go#L28) to 5. As I was implementing this functional way, I was a bit paranoid that I might have different number of Cas-engines support in functions, i.e. some volumes might not show up unless a flag is passed or otherwise, so I pulled up some hacks to stop this in the PR phase itself. You can run go tests locally with `go test ./...`

pkg/volume/local_hostpath.go
Abhishek-kumar09: Done

pkg/client/k8s.go
Abhishek-kumar09: Yeah, sounds good. There are some other functions also that can be optimized in the same way.

pkg/client/k8s.go
Abhishek-kumar09: Applied this :smiley:

pkg/client/k8s.go
```diff
@@ -743,3 +743,19 @@ func (k K8sClient) GetPods(labelSelector string, fieldSelector string, namespace
...
+
+// GetLocalPvDeployment returns the deployment-list with a specific
+// label selector query
+func (k K8sClient) GetLocalPvDeployment(labelSelector string) (*appsv1.DeploymentList, error) {
+	if pv, err := k.K8sCS.AppsV1().Deployments("").List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labelSelector,
+	}); err == nil && len(pv.Items) >= 1 {
+		return pv, nil
+	} else {
+		return nil, fmt.Errorf("got 0 local-volumes with the label openebs.io/component-name=%s", labelSelector)
```
vharsh: this error message needs an update now

pkg/volume/volume.go
vharsh: @Abhishek-kumar09 This change isn't required, as the namespace of LocalHostPath resource can be guessed correctly via the `nsMap` variable

pkg/client/k8s.go
```diff
@@ -743,3 +743,19 @@ func (k K8sClient) GetPods(labelSelector string, fieldSelector string, namespace
...
 	return pods, nil
 }
+
+/*
+	LOCAL VOLUMES SPECIFIC METHODS
+*/
+
+// GetLocalPvDeployment returns the deployment-list with a specific
+// label selector query
+func (k K8sClient) GetLocalPvDeployment(labelSelector string) (*appsv1.DeploymentList, error) {
```
vharsh: this function can be renamed to `GetDeployments` instead, do update the name in the comment too.

pkg/util/constant.go
vharsh: I'll confirm about this in an hour for sure, so that release can be triggered timely.

vharsh (commented):
last set of comments 😅

pkg/client/k8s.go
Abhishek-kumar09: Sorry, My mistake. Corrected.

pkg/util/constant.go
```diff
@@ -33,10 +33,12 @@ const (
...
 	// LVMCasType cas type name
 	LVMCasType = "localpv-lvm"
-	// LocalHostpathCasType cas type name
-	LocalHostpathCasType = "localpv-hostpath"
+	// LocalPvHostpathCasType cas type name
+	LocalPvHostpathCasType = "localpv-hostpath"
 	// LocalDeviceCasType cas type name
 	LocalDeviceCasType = "localpv-device"
+	// Local-hostpath cas type name
+	LocalHostpathCasType = "local-hostpath"
```
vharsh, line 41:
```suggestion
```

This can be removed, I've discussed this with Abhinandan, we are sort of in the fence when it comes to naming things right now, but we are positive that removing this will not change anything.

pkg/util/constant.go
```diff
@@ -33,10 +33,12 @@ const (
 	JivaCasType = "jiva"
 	// LVMCasType cas type name
 	LVMCasType = "localpv-lvm"
-	// LocalHostpathCasType cas type name
-	LocalHostpathCasType = "localpv-hostpath"
+	// LocalPvHostpathCasType cas type name
+	LocalPvHostpathCasType = "localpv-hostpath"
```
vharsh, line 37:
```suggestion
	LocalPvHostpathCasType = "local-hostpath"
```

Update this constant to local-hostpath, and change the variables elsewhere accordingly.

pkg/util/constant.go
vharsh: @kmova Both the constants are required, because the LocalPV-hostpath PV has this label `openebs.io/cas-type=local-hostpath`

pkg/volume/local_hostpath.go
```diff
@@ -0,0 +1,94 @@
...
+func GetLocalHostpath(c *client.K8sClient, pvList *corev1.PersistentVolumeList, openebsNS string) ([]metav1.TableRow, error) {
+	var rows []metav1.TableRow
+	for _, pv := range pvList.Items {
+		// Ignore all the other volumes that is not of cas-type local-hostpath
+		if util.GetCasTypeFromPV(&pv) != util.LocalHostpathCasType {
+			continue
+		}
+
+		name := pv.Name
+		capacity := pv.Spec.Capacity.Storage()
```
Abhinandan-Purkait, line 53: @vharsh Do we need to humanize this unit?

vharsh: No, it's already humanized if it's coming from corev1.PersistentVolume

pkg/util/constant.go
vharsh: @Abhishek-kumar09 I've confirmed, you'd be needing both of them, the cas-type name is `localpv-hostpath` but the Dynamic-LocalPV-provisioner adds the label `openebs.io/cas-type=local-hostpath`(something which would be fixed at a later stage after checking if this change doesn't break things like upgrade.

vharsh:
merging this for release

Abhinandan-Purkait (approved):

PR #113, opened 2021-09-14
fixed incorrect display message in version issue #111

### LINKED ISSUE #111: incorrect display message in version subcommand, when using custom build.
When using a custom built openebs ctl, the version command shows the following:

```
You are using an older version (dev) of cli, latest available version is: v0.3.1
```

When using the dev version. The above message should change to:
```
You are using development version of cli, latest released version is: v0.3.1
```

Amishakumari544:
Fixes #111

### DIFF
--- a/cmd/version/version.go
+++ b/cmd/version/version.go
@@ -117,14 +117,14 @@ func checkForLatestVersion(currVersion string) {
 		return
 	}
 
-	currversion := data["spec"].(map[string]interface{})["version"].(string)
+	
 	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
-	if currversion == "dev" {
+	if !isLatestVersion(currVersion, latestVersion) {
 		fmt.Println()
-		fmt.Printf("You are using development version of cli, latest released version is: %s", currversion)
-	} else if !isLatestVersion(currVersion, latestVersion) {
-		// The seperator for the error print
-		fmt.Println()
+		if currVersion == "dev" {
+			fmt.Printf("You are using development version of cli, latest released version is: %s", latestVersion)
+			return
+		}
 		fmt.Printf("You are using an older version (%s) of cli, latest available version is: %s", currVersion, latestVersion)
 	}
 }
--- a/cmd/version/version.go
+++ b/cmd/version/version.go
@@ -117,8 +117,12 @@ func checkForLatestVersion(currVersion string) {
 		return
 	}
 
+	currversion := data["spec"].(map[string]interface{})["version"].(string)
 	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
-	if !isLatestVersion(currVersion, latestVersion) {
+	if currversion == "dev" {
+		fmt.Println()
+		fmt.Printf("You are using development version of cli, latest released version is: %s", currversion)
+	} else if !isLatestVersion(currVersion, latestVersion) {
 		// The seperator for the error print
 		fmt.Println()
 		fmt.Printf("You are using an older version (%s) of cli, latest available version is: %s", currVersion, latestVersion)
--- a/go.mod
+++ b/go.mod
@@ -9,10 +9,12 @@ require (
 	github.com/openebs/lvm-localpv v0.6.0
 	github.com/openebs/zfs-localpv v1.8.0
 	github.com/pkg/errors v0.9.1
+	github.com/ramya-rao-a/go-outline v0.0.0-20210608161538-9736a4bde949 // indirect
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.0
 	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 // indirect
+	golang.org/x/tools v0.1.6 // indirect
 	k8s.io/api v0.20.2
 	k8s.io/apimachinery v0.20.2
 	k8s.io/cli-runtime v0.20.0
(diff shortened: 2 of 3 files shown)


Amishakumari544:
Screenshot (47)


**Println** is giving warning so, I replaced this with  **Printf**

cmd/version/version.go
```diff
@@ -117,8 +117,12 @@ func checkForLatestVersion(currVersion string) {
 		return
 	}
 
+	currversion := data["spec"].(map[string]interface{})["version"].(string)
```
Abhishek-kumar09: The [data](https://raw.githubusercontent.com/kubernetes-sigs/krew-index/master/plugins/openebs.yaml) contains latest details about the openebsctl deployment distributed via krew.
We are already getting `currVersion` from the function argument. 
```suggestion

```

cmd/version/version.go
```diff
@@ -117,8 +117,12 @@ func checkForLatestVersion(currVersion string) {
...
 	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
-	if !isLatestVersion(currVersion, latestVersion) {
+	if currversion == "dev" {
+		fmt.Println()
+		fmt.Printf("You are using development version of cli, latest released version is: %s", currversion)
+	} else if !isLatestVersion(currVersion, latestVersion) {
```
Abhishek-kumar09: Can we do something like:
```go
if !islatestVersion(currVersion, latestVersion) {
  fmt.Println()
  if currVersion == "dev" {
    fmt.Printf("You are using development version of cli, latest released version is: %s", latestVersion)
    return
  }
  fmt.Printf("You are using an older version (%s) of cli, latest available version is: %s", currVersion, latestVersion)
}
```

Abhishek-kumar09 (changes requested):
Thanks for the fix @Amishakumari544 :smiley: 
Some minors

Abhishek-kumar09:
> Screenshot (47)
> 
> **Println** is giving warning so, I replaced this with **Printf**

Yes actually Println doesn't support string Formatting, thus you can't use, `%s` with it.
Instead, if required you can use Println like string concatenation, breaking string with `+` signs. like

```go
greet := "world"
fmt.Println("hello" + greet)
 ```

File: cmd/version/version.go (hunk 1 above)
Comment (Amishakumari544) [reply cmd/version/version.go]: ok then I am removing this.

File: cmd/version/version.go (hunk 2 above)
Comment (kmova) [reply cmd/version/version.go]: Good one!

File: cmd/version/version.go [hunk 3]
```diff
@@ -117,8 +117,12 @@ func checkForLatestVersion(currVersion string) {
 		return
 	}
 
+	currversion := data["spec"].(map[string]interface{})["version"].(string)
 	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
-	if !isLatestVersion(currVersion, latestVersion) {
+	if currversion == "dev" {
+		fmt.Println()
+		fmt.Printf("You are using development version of cli, latest released version is: %s", currversion)
```
avishnu: should be latestVersion instead of currVersion.

go.mod
```diff
@@ -9,10 +9,12 @@ require (
 	github.com/openebs/lvm-localpv v0.6.0
 	github.com/openebs/zfs-localpv v1.8.0
 	github.com/pkg/errors v0.9.1
+	github.com/ramya-rao-a/go-outline v0.0.0-20210608161538-9736a4bde949 // indirect
```
vharsh, line 12: This file should not have been changed.

go.sum
```diff
@@ -383,6 +383,8 @@ github.com/prometheus/procfs v0.0.2/go.mod h1:TjEm7ze935MbeOT/UhFTIMYKhuLP4wbCsT
 github.com/prometheus/procfs v0.1.3/go.mod h1:lV6e/gmhEcM9IjHGsFOCxxuZ+z1YqCvr4OA4YeYWdaU=
 github.com/prometheus/procfs v0.2.0/go.mod h1:lV6e/gmhEcM9IjHGsFOCxxuZ+z1YqCvr4OA4YeYWdaU=
 github.com/prometheus/tsdb v0.7.1/go.mod h1:qhTCs0VvXwvX/y3TZrWD7rabWM+ijKTux40TwIPHuXU=
+github.com/ramya-rao-a/go-outline v0.0.0-20210608161538-9736a4bde949 h1:iaD+iVf9xGfajsJp+zYrg9Lrk6gMJ6/hZHO4cYq5D5o=
```
vharsh, line 386: Same, please remove the changes from this file.

cmd/version/version.go
```diff
@@ -117,10 +117,14 @@ func checkForLatestVersion(currVersion string) {
 		return
 	}
 
+	
 	latestVersion := data["spec"].(map[string]interface{})["version"].(string)
 	if !isLatestVersion(currVersion, latestVersion) {
-		// The seperator for the error print
 		fmt.Println()
+		if currVersion == "dev" {
+			fmt.Printf("You are using development version of cli, latest released version is: %s", latestVersion)
```
vharsh, line 125:
```suggestion
			fmt.Println("You are using development version of cli, latest released version is: " + latestVersion)
```

cmd/version/version.go
```diff
@@ -117,10 +117,14 @@ func checkForLatestVersion(currVersion string) {
...
+			return
+		}
 		fmt.Printf("You are using an older version (%s) of cli, latest available version is: %s", currVersion, latestVersion)
```
vharsh, line 128: Please convert it to `fmt.Println` so that a newline(the correct one depending upon the system) is added.
The newline is a confusing character across operating systems and sometimes it's best left for the libraries to deal with them correctly.
Ref: https://stackoverflow.com/a/3720674

go.mod
vharsh: You might need to rebase your code against `develop` branch, this library isn't required by any change you've done IMO.

vharsh (changes requested):
please don't commit the binary file in the PR

cmd/version/version.go
vharsh: > Println is giving warning so, I replaced this with Printf

That's because Println had `%s`, as it was a format specifier or a placeholder for a string, this one shouldn't give warnings.

PR #120, opened 2021-09-26
refractor: Wrap NewK8sClient for error handling in client init

Abhishek-kumar09:
In the case of wrong kubeconfig, or no kubeconfig or no kubernetes the client initialisation throws error which are not handled by the current code at multiple places resulting in panic by CLI. 
- [x] Wrapped the client initialization with error-handling implicitly. 
- [x] Made namespace as optional argument, so that if we don't want to provide namespace and want to search in all namespaces we can omit passing empty string in functional arguments

### DIFF
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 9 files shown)


pkg/client/k8s.go
```diff
@@ -72,10 +73,33 @@ type K8sClient struct {
 	CLIENT CREATION METHODS AND RELATED OPERATIONS
 */
 
-// NewK8sClient creates a new K8sClient
+// A wrapper around newK8sClient to handle errors in creating
```
vharsh: The comment should begin like `NewK8sClient ...`

This is a golint error.

vharsh (approved):
minor suggestion, great work on avoiding a change in function names so that other code won't break 🥳 🎉, fewer merge conflicts 👍🏼

pkg/client/k8s.go
Abhishek-kumar09: Done

vharsh (approved):
lgtm

PR #123, opened 2021-10-05
Add support for creating a CSPC template

### LINKED ISSUE #58: Support creating an cStor Pool Cluster (CSPC) template
Setting up cStor storage is a cumbersome process - that requires users to fetch the appropriate BDs, then create an CSPC yaml file to apply. This can lead to many mis-configurations. 

It will be nice to have a cli command that would help user automatically create an CSPC yaml file that can be reviewed by the user and applied into the cluster. 

One way to implement this could be:

`kubectl openebs template cspc --nodes=node1,node2,node3 [--number-of-devices=2 --raidtype=mirror]`

- provide a generic `template` subcommand that can be extended for other usecases
- `cspc` can be the first template 
- `nodes` is a mandatory argument - can be one or more node names (obtained via `kubectl get nodes`). 
- optional: ask for number of block devices to be used (default to 1)
- optional: ask for raidtype (default to striped)
- Other optional parameters that can be taken up are - type of block device, min capacity  of block device)

vharsh:
# Summary

For a k8s cluster which has CStor installed, this CLI takes some arguments and generates a Pool-Cluster YAML to help user create a CStor pool of a certain kind via flags

fixes: #58 

```bash
$ kubectl openebs generate cspc --nodes=shubham
apiVersion: cstor.openebs.io/v1
kind: CStorPoolCluster
metadata:
  creationTimestamp: null
  generateName: cstor
  namespace: openebs
spec:
  pools:
  - dataRaidGroups:
    - blockDevices:
      # /var/openebs/sparse/0-ndm-sparse.img  10GB
      - blockDeviceName: sparse-6b277da87b7487e501c03ea0001d6d92
    nodeSelector:
      kubernetes.io/hostname: shubham
    poolConfig:
      dataRaidGroupType: stripe
```

## Pending items
- [ ] Unit testing
- [x] Implementing `makePools` for other poolTypes
- [ ] Adding capacity restriction for the pools

## Supported poolTypes

- [x] stripe
- [x] mirror
- [ ] raidz
- [ ] raidz2

### DIFF
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -52,7 +52,8 @@ func NewCmdGenerateStorage() *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
 		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
 	cmd.MarkPersistentFlagRequired("nodes")
-	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "--number-of-devices=2")
-	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe", "--raidtype=mirrored")
+	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node")
+	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe",
+		"allowed RAID configuration such as, stripe, mirror, raid, raidz2")
 	return cmd
 }
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -51,7 +51,7 @@ func NewCmdGenerateStorage() *cobra.Command {
 	}
 	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
 		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
-	cmd.MarkPersistentFlagRequired("nodes")
+	_ = cmd.MarkPersistentFlagRequired("nodes")
 	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node")
 	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe",
 		"allowed RAID configuration such as, stripe, mirror, raid, raidz2")
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2020-2021 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generate
+
+import (
+	"strings"
+
+	"github.com/openebs/openebsctl/pkg/generate"
+	"github.com/openebs/openebsctl/pkg/util"
+	"github.com/spf13/cobra"
+)
+
+// NewCmdGenerate provides options for generating
+func NewCmdGenerate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:       "generate",
+		ValidArgs: []string{"cspc"},
+		Short:     "Generate one or more OpenEBS resources based on flags",
+	}
+	cmd.AddCommand(NewCmdGenerateStorage())
+	return cmd
+}
+
+func NewCmdGenerateStorage() *cobra.Command {
+	var nodes, raidType string
+	var devices int
+	cmd := &cobra.Command{
+		Use:   "cspc",
+		Short: "Displays status information about Volume(s)",
+		Run: func(cmd *cobra.Command, args []string) {
+			node, _ := cmd.Flags().GetString("nodes")
+			devs, _ := cmd.Flags().GetInt("number-of-devices")
+			raid, _ := cmd.Flags().GetString("raidtype")
+			Nodes := strings.Split(node, ",")
+			util.CheckErr(generate.Pool(Nodes, devs, raid), util.Fatal)
+		},
+	}
+	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
+		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
+	cmd.MarkPersistentFlagRequired("nodes")
+	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "--number-of-devices=2")
+	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe", "--raidtype=mirrored")
+	return cmd
+}
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -17,6 +17,7 @@ limitations under the License.
 package generate
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/openebs/openebsctl/pkg/generate"
@@ -28,32 +29,62 @@ import (
 func NewCmdGenerate() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:       "generate",
+		Short:     "Generate one or more OpenEBS resource like cspc",
 		ValidArgs: []string{"cspc"},
-		Short:     "Generate one or more OpenEBS resources based on flags",
 	}
-	cmd.AddCommand(NewCmdGenerateStorage())
+	cmd.AddCommand(NewCmdGenerateCStorStoragePoolCluster())
 	return cmd
 }
 
-func NewCmdGenerateStorage() *cobra.Command {
+// NewCmdGenerateCStorStoragePoolCluster provides options for generating cspc
+// NOTE: When other custom resources need to be generated, the function
+// should be renamed appropriately, as of now it made no sense to generically
+// state pools when other pools aren't supported.
+func NewCmdGenerateCStorStoragePoolCluster() *cobra.Command {
 	var nodes, raidType string
 	var devices int
 	cmd := &cobra.Command{
-		Use:   "cspc",
-		Short: "Displays status information about Volume(s)",
+		Use:     "cspc",
+		Short:   "Generates cspc resources YAML/configuration which can be used to provision cStor storage pool clusters",
 		Run: func(cmd *cobra.Command, args []string) {
 			node, _ := cmd.Flags().GetString("nodes")
-			devs, _ := cmd.Flags().GetInt("number-of-devices")
 			raid, _ := cmd.Flags().GetString("raidtype")
-			Nodes := strings.Split(node, ",")
-			util.CheckErr(generate.Pool(Nodes, devs, raid), util.Fatal)
+			devs := numDevices(cmd)
+			nodeList := strings.Split(node, ",")
+			util.CheckErr(generate.CSPC(nodeList, devs, raid), util.Fatal)
 		},
 	}
 	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
 		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
 	_ = cmd.MarkPersistentFlagRequired("nodes")
-	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node")
 	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe",
 		"allowed RAID configuration such as, stripe, mirror, raid, raidz2")
+	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node, selects default based on raid-type")
 	return cmd
+}
+
+// numDevices figures out the number of devices based on the raid type
+func numDevices(cmd *cobra.Command) int {
+	// if number-of-devices is not set, set it to appropriate value
+	if !cmd.Flag("number-of-devices").Changed {
+		var devCount = map[string]int{
+			"stripe": 1,
+			"mirror": 2,
+			"raidz":  3,
+			"raidz2": 4}
+		switch cmd.Flag("raidtype").Value.String() {
+		case "stripe", "mirror", "raidz", "raidz2":
+			c := devCount[cmd.Flag("raidtype").Value.String()]
+			err := cmd.Flags().Set("number-of-devices", strconv.Itoa(c))
+			if err != nil {
+				return 1
+			}
+			return c
+		}
+	} else {
+		d, _ := cmd.Flags().GetInt("number-of-devices")
+		return d
+	}
+	// setting default value to 1
+	return 1
 }
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.0
+	github.com/stretchr/testify v1.6.1
 	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
 	golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -506,7 +506,7 @@ func (k K8sClient) GetNodes(nodes []string, label, field string) (*corev1.NodeLi
 	if len(nodes) == 0 {
 		return n, err
 	}
-	// 2. Put them in a map[string]*corev1.Node
+	// 2. Put them in a map[string]corev1.Node
 	nodeMap := make(map[string]corev1.Node)
 	for _, item := range n.Items {
 		nodeMap[item.Name] = item
--- a/pkg/client/k8s.go
+++ b/pkg/client/k8s.go
@@ -267,18 +267,6 @@ func (k K8sClient) GetAllPods(namespace string) (*corev1.PodList, error) {
 	return pods, nil
 }
 
-// GetDeploymentList returns the deployment-list with a specific
-// label selector query
-func (k K8sClient) GetDeploymentList(labelSelector string) (*appsv1.DeploymentList, error) {
-	if pv, err := k.K8sCS.AppsV1().Deployments("").List(context.TODO(), metav1.ListOptions{
-		LabelSelector: labelSelector,
-	}); err == nil && len(pv.Items) >= 1 {
-		return pv, nil
-	} else {
-		return nil, fmt.Errorf("got 0 deployments with label-Selector as %s", labelSelector)
-	}
-}
-
 // GetSC returns a StorageClass object using the scName passed.
 func (k K8sClient) GetSC(scName string) (*v1.StorageClass, error) {
 	sc, err := k.K8sCS.StorageV1().StorageClasses().Get(context.TODO(), scName, metav1.GetOptions{})
@@ -492,4 +480,45 @@ func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
 // Delete a Batch Job by name
 func (k K8sClient) DeleteBatchJob(name string, namespace string) error {
 	return k.K8sCS.BatchV1().Jobs(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+}
+
+/*
+	LOCAL VOLUMES SPECIFIC METHODS
+*/
+
+// GetDeploymentList returns the deployment-list with a specific
+// label selector query
+func (k K8sClient) GetDeploymentList(labelSelector string) (*appsv1.DeploymentList, error) {
+	if pv, err := k.K8sCS.AppsV1().Deployments("").List(context.TODO(), metav1.ListOptions{
+		LabelSelector: labelSelector,
+	}); err == nil && len(pv.Items) >= 1 {
+		return pv, nil
+	}
+	return nil, fmt.Errorf("got 0 deployments with label-Selector as %s", labelSelector)
+}
+
+// GetNodes returns a list of nodes with the name of nodes
+func (k K8sClient) GetNodes(nodes []string, label, field string) (*corev1.NodeList, error) {
+	// 1. Get all nodes
+	n, err := k.K8sCS.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+		LabelSelector: label,
+		FieldSelector: field})
+	if len(nodes) == 0 {
+		return n, err
+	}
+	// 2. Put them in a map[string]*corev1.Node
+	nodeMap := make(map[string]corev1.Node)
+	for _, item := range n.Items {
+		nodeMap[item.Name] = item
+	}
+	// 3. Get the nodes by the name nodes
+	var items []corev1.Node
+	for _, name := range nodes {
+		if _, ok := nodeMap[name]; ok {
+			items = append(items, nodeMap[name])
+		}
+	}
+	return &corev1.NodeList{
+		Items: items,
+	}, nil
 }
--- a/pkg/generate/cspc.go
+++ b/pkg/generate/cspc.go
@@ -33,19 +33,18 @@ import (
 func isPoolTypeValid(raid string) bool {
 	if raid == "stripe" || raid == "mirror" || raid == "raidz" || raid == "raidz2" {
 		return true
-	} else {
-		return false
 	}
+	return false
 }
 
-// Pool calls the generate routine for different cas-types
-func Pool(nodes []string, devs int, raid string) error {
+// CSPC calls the generate routine for different cas-types
+func CSPC(nodes []string, devs int, raid string) error {
 	c := client.NewK8sClient()
 	if !isPoolTypeValid(strings.ToLower(raid)) {
 		// TODO: Use the well defined pool constant types from openebs/api when added there
 		return fmt.Errorf("invalid pool type %s", raid)
 	}
-	_, str, err := CSPC(c, nodes, devs, strings.ToLower(raid))
+	_, str, err := cspc(c, nodes, devs, strings.ToLower(raid))
 	if err != nil {
 		return err
 	}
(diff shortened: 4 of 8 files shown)


pkg/client/k8s.go
```diff
@@ -493,3 +481,44 @@ func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
...
 	return k.K8sCS.BatchV1().Jobs(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
 }
+
+/*
+	LOCAL VOLUMES SPECIFIC METHODS
+*/
+
+// GetDeploymentList returns the deployment-list with a specific
+// label selector query
+func (k K8sClient) GetDeploymentList(labelSelector string) (*appsv1.DeploymentList, error) {
```
vharsh, line 491: This function got added via rebase

pkg/generate/cspc.go
```diff
@@ -0,0 +1,215 @@
...
+	cstorv1 "github.com/openebs/api/v2/pkg/apis/cstor/v1"
+	"github.com/openebs/api/v2/pkg/apis/openebs.io/v1alpha1"
+	"github.com/openebs/openebsctl/pkg/client"
+	"github.com/openebs/openebsctl/pkg/util"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// isPoolValid checks if a CStor pool is valid
+func isPoolTypeValid(raid string) bool {
+	if raid == "stripe" || raid == "mirror" || raid == "raidz" || raid == "raidz2" {
```
vharsh, line 34: please clarify if `mirror` or `mirrored` both are accepted terms and can be used interchangeably

pkg/generate/cspc.go
```diff
@@ -0,0 +1,215 @@
...
+	_, str, err := CSPC(c, nodes, devs, strings.ToLower(raid))
+	if err != nil {
+		return err
+	}
+	fmt.Println(str)
+	return nil
+}
+
+// CSPC takes eligible nodes, number of devices and poolType to create a pool cluster template
+func CSPC(c *client.K8sClient, nodes []string, devs int, poolType string) (*cstorv1.CStorPoolCluster, string, error) {
```
vharsh: I've added the structure, YAML string, error as return values to help while writing tests, the structure isn't used by callers, but it can be used later to augment some functionality later

pkg/generate/cspc.go
```diff
@@ -0,0 +1,215 @@
...
+	// 0. Figure out the OPENEBS_NAMESPACE for CStor
+	cstorNS, err := c.GetOpenEBSNamespace(util.CstorCasType)
+	// assume CSTOR's OPENEBS_NAMESPACE has all the relevant blockdevices
+	c.Ns = cstorNS
+	if err != nil {
+		return nil, "", fmt.Errorf("cannot find an active cstor installation")
```
vharsh: Please see if this error message is appropriate, my idea is, if the CSI driver labeled pod is missing and is not in a Running state, there's no active cstor installation.

pkg/generate/cspc.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	// 1. Validate nodes & poolType, fetch disks
+	nodeList, err := c.GetNodes(nodes, "", "")
```
vharsh, line 65: Should only `status.Phase=Running` nodes be listed?

pkg/generate/cspc_test.go
```diff
@@ -0,0 +1,143 @@
...
+	cstorv1 "github.com/openebs/api/v2/pkg/apis/cstor/v1"
+	cstorfake "github.com/openebs/api/v2/pkg/client/clientset/versioned/fake"
+	"github.com/openebs/openebsctl/pkg/client"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+
+
+func TestCSPC(t *testing.T) {
```
vharsh, line 32: This is more like an integration test as it depends on >1 function, I'll refactor this later, such that an instance of `makePools` or `filterBDs` are called, this will also help expand the generation of other type of pool YAMLs or other similar resources such as JVP, SC(CStor & Jiva), etc.

pkg/generate/testdata_test.go
```diff
@@ -0,0 +1,197 @@
...
+      - blockDeviceName: bd1-n2
+      # /dev/sda  1.0GiB
+      - blockDeviceName: bd2-n2
+    nodeSelector:
+      kubernetes.io/hostname: node2
+    poolConfig:
+      dataRaidGroupType: mirror
+  - dataRaidGroups:
+    - blockDevices:
+      # /dev/sda  1.0GiB
```
vharsh, line 204: The size is displayed in a weird way now, I'll fix it sometime soon.

pkg/generate/cspc.go
```diff
@@ -0,0 +1,215 @@
...
+	for _, bd := range bds.Items {
+		// an eligible blockdevice is in active+unclaimed state and lacks a file-system
+		if bd.Status.State == v1alpha1.BlockDeviceActive &&
+			bd.Status.ClaimState == v1alpha1.BlockDeviceUnclaimed &&
+			bd.Spec.FileSystem.Type == "" &&
+			// BD's capacity >=64 MiB
+			bd.Spec.Capacity.Storage >= 67110000 {
+			final = append(final, bd)
+		}
+	}
```
vharsh, line 237: [CSTOR] Is this criteria sufficient or it should check the NDM config for `auto` & `manual` modes so that it can reject sparse files or something else?

pkg/generate/testdata_test.go
```diff
@@ -0,0 +1,197 @@
...
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n1"}, {BlockDeviceName: "bd2-n1"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}},
+		{NodeSelector: map[string]string{"kubernetes.io/hostname": "node2"},
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n2"}, {BlockDeviceName: "bd2-n2"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}},
+		{NodeSelector: map[string]string{"kubernetes.io/hostname": "node3"},
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n3"}, {BlockDeviceName: "bd2-n3"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}}}}}
+
+var mirrorCSPCstr = `apiVersion: cstor.openebs.io/v1
```
vharsh, line 174: [CSTOR] Does this `mirror` configuration look good in case I've got two disks of 1GiB on each of 3 worker nodes with hostnames `, what more detail should I add in comments.
Some good suggestions can be

1. Actual size available to applications depending upon pool type.
2. PoolType features, when should one go for mirror, stripe, raidz, raidz2, etc.

pkg/generate/testdata_test.go
```diff
@@ -0,0 +1,197 @@
...
+      # /dev/sda  1.0GiB
+      - blockDeviceName: bd1-n1
+      # /dev/sda  1.0GiB
+      - blockDeviceName: bd2-n1
+    nodeSelector:
+      kubernetes.io/hostname: node1
+    poolConfig:
+      dataRaidGroupType: mirror
+  - dataRaidGroups:
+    - blockDevices:
```
vharsh, line 193: [CSTOR] The number of devices in a CSPIbd should be a multiple of two and there can be any number of `dataRaidGroups` 

Took this suggestion from [webhook validation](https://github.com/openebs/cstor-operators/blob/develop/pkg/webhook/cspc.go#L302), thanks @mittachaitu for pointing to the actual codebase.

cmd/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+			Nodes := strings.Split(node, ",")
+			util.CheckErr(generate.Pool(Nodes, devs, raid), util.Fatal)
+		},
+	}
+	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
+		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
+	_ = cmd.MarkPersistentFlagRequired("nodes")
+	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node")
+	cmd.PersistentFlags().StringVarP(&raidType, "raidtype", "", "stripe",
+		"allowed RAID configuration such as, stripe, mirror, raid, raidz2")
```
vharsh, line 61: These are the exact poolType phrases/terms in flags & everywhere else.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.0
+	github.com/stretchr/testify v1.6.1
```
vharsh, line 16: for nested structures, JSONs, YAMLs, this seems to be the best library possible; other similar-ish lightweight libraries are [is](https://pkg.go.dev/github.com/matryer/is#section-readme) and a cool library which we might like better [gocmp](https://github.com/google/go-cmp/).

# Reasons for preferring go-cmp for comparing structs

1. [Cleaner one line-ish diffs of structs](https://blog.artwolf.in/a?ID=2e0cd2a0-3dfd-4dbc-a87d-c40647552b9a)
2. [Mentioned by Dave Cheney in testing blogs](https://dave.cheney.net/tag/testing)

pkg/generate/cspc.go
mittachaitu: The supported type is "mirror"... refer [apis](https://github.com/openebs/api/blob/ecd9a8a61d3ef0b2e3cc7a9c2d304eea9b5ce6d2/pkg/apis/cstor/v1/cstorpoolcluster.go#L25-L37) if required any clarification.

cmd/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+	cmd := &cobra.Command{
+		Use:       "generate",
+		ValidArgs: []string{"cspc"},
+		Short:     "Generate one or more OpenEBS resources based on flags",
+	}
+	cmd.AddCommand(NewCmdGenerateStorage())
+	return cmd
+}
+
+func NewCmdGenerateStorage() *cobra.Command {
```
mittachaitu: nit: Good to have a function name more specific to CSPC/cStor pool instead of a generic name(in the future we can accommodate mayastor pools as well)
```suggestion
func NewCmdGenerateCstorStoragePool() *cobra.Command {
```

cmd/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+	var nodes, raidType string
+	var devices int
+	cmd := &cobra.Command{
+		Use:   "cspc",
+		Short: "Displays status information about Volume(s)",
```
mittachaitu: Generates CSPC resources YAML/configuration which can be used to provision cStor storagepools

cmd/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+			devs, _ := cmd.Flags().GetInt("number-of-devices")
+			raid, _ := cmd.Flags().GetString("raidtype")
+			Nodes := strings.Split(node, ",")
+			util.CheckErr(generate.Pool(Nodes, devs, raid), util.Fatal)
+		},
+	}
+	cmd.PersistentFlags().StringVarP(&nodes, "nodes", "", "",
+		"comma separated set of nodes for pool creation --nodes=node1,node2,node3,node4")
+	_ = cmd.MarkPersistentFlagRequired("nodes")
+	cmd.PersistentFlags().IntVar(&devices, "number-of-devices", 1, "number of devices per node")
```
mittachaitu: Can we default it to no.of devices based on raidType? It will be helpful for users not to go through documentation

cmd/generate/generate.go
```diff
@@ -0,0 +1,59 @@
...
+	var nodes, raidType string
+	var devices int
+	cmd := &cobra.Command{
+		Use:   "cspc",
+		Short: "Displays status information about Volume(s)",
+		Run: func(cmd *cobra.Command, args []string) {
+			node, _ := cmd.Flags().GetString("nodes")
+			devs, _ := cmd.Flags().GetInt("number-of-devices")
+			raid, _ := cmd.Flags().GetString("raidtype")
+			Nodes := strings.Split(node, ",")
```
mittachaitu: nit:
```suggestion
			nodeNameList := strings.Split(node, ",")
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+func isPoolTypeValid(raid string) bool {
+	if raid == "stripe" || raid == "mirror" || raid == "raidz" || raid == "raidz2" {
+		return true
+	} else {
+		return false
+	}
+}
+
+// Pool calls the generate routine for different cas-types
+func Pool(nodes []string, devs int, raid string) error {
```
mittachaitu: nit: Good to have function name related to CSPC/cStor pool
```suggestion
func CStorPool(nodes []string, devs int, raid string) error {
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// isPoolValid checks if a CStor pool is valid
+func isPoolTypeValid(raid string) bool {
+	if raid == "stripe" || raid == "mirror" || raid == "raidz" || raid == "raidz2" {
+		return true
+	} else {
+		return false
+	}
```
mittachaitu:
```suggestion
	}
	return false
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// CSPC takes eligible nodes, number of devices and poolType to create a pool cluster template
+func CSPC(c *client.K8sClient, nodes []string, devs int, poolType string) (*cstorv1.CStorPoolCluster, string, error) {
+	// 0. Figure out the OPENEBS_NAMESPACE for CStor
+	cstorNS, err := c.GetOpenEBSNamespace(util.CstorCasType)
+	// assume CSTOR's OPENEBS_NAMESPACE has all the relevant blockdevices
+	c.Ns = cstorNS
+	if err != nil {
+		return nil, "", fmt.Errorf("cannot find an active cstor installation")
```
mittachaitu:
```suggestion
		return nil, "", fmt.Errorf("unable to determine the cStor namespace error: %v", err)
```
We are swallowing the original error in [c.GetOpenEBSNamespace](https://github.com/openebs/openebsctl/blob/9b279df44ee1c8c9d69950b08015690c8a0944b2/pkg/client/k8s.go#L188) good to return different error based on err variable... Something like below:


```go
func (k K8sClient) GetOpenEBSNamespace(casType string) (string, error) {
	pods, err := k.K8sCS.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{FieldSelector: "status.phase=Running", LabelSelector: fmt.Sprintf("openebs.io/component-name=%s", util.CasTypeAndComponentNameMap[strings.ToLower(casType)])})
	if err != nil {
		return "", errors.Wrapf(err, "unable to determine openebs namespace")
	}
        if len(pods.Items) == 0 {
                return "", errors.New("unable to find cStor related artifacts")
        }
	return pods.Items[0].Namespace, nil
}
```

pkg/client/k8s.go
```diff
@@ -493,3 +481,44 @@ func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
...
+// GetNodes returns a list of nodes with the name of nodes
+func (k K8sClient) GetNodes(nodes []string, label, field string) (*corev1.NodeList, error) {
+	// 1. Get all nodes
+	n, err := k.K8sCS.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{
+		LabelSelector: label,
+		FieldSelector: field})
+	if len(nodes) == 0 {
+		return n, err
+	}
+	// 2. Put them in a map[string]*corev1.Node
```
mittachaitu:
```suggestion
	// 2. Put them in a map[string]corev1.Node
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	// 1. Validate nodes & poolType, fetch disks
+	nodeList, err := c.GetNodes(nodes, "", "")
+	if err != nil {
+		return nil, "", fmt.Errorf("(server error) unable to fetch nodes %s", err)
```
mittachaitu:
```suggestion
		return nil, "", fmt.Errorf("(server error) unable to fetch node information %s", err)
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	if len(nodeList.Items) != len(nodes) {
+		return nil, "", fmt.Errorf("not all worker nodes are available for provisioning a CSPC")
+	}
+	// 2. Fetch BD's from the eligible/valid nodes
+	bds, err := c.GetBDs(nil, "kubernetes.io/hostname in ("+strings.Join(nodes, ",")+")")
```
mittachaitu: This will not work in the case of EKS based clusters since in EKS setup nodeName is different from the `kubernetes.io/hostname` value. So better to get the hostNames for given nodeNames and then query for bds

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil || len(bds.Items) == 0 {
+		return nil, "", fmt.Errorf("no blockdevices found in %s nodes", nodes)
```
mittachaitu: Let's also log about the error and do we need to change the format specifier to %v since nodes is a string slice

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+					DataRaidGroupType: string(cstorv1.PoolStriped),
+				},
+			})
+		}
+		return &spec, nil
+	} else if poolType == string(cstorv1.PoolMirrored) {
+		if nDevices%2 != 0 {
+			return nil, fmt.Errorf("mirrored pool requires multiples of two block device")
+		}
+		for node, bds := range bd {
```
mittachaitu:
```suggestion
		for hostName, bds := range bd {
```

pkg/generate/cspc.go
```diff
@@ -0,0 +1,214 @@
...
+			var raids []cstorv1.CStorPoolInstanceBlockDevice
+			// add all BDs to a CSPCs CSPI spec
+			for d := 0; d < nDevices; d++ {
+				raids = append(raids, cstorv1.CStorPoolInstanceBlockDevice{BlockDeviceName: bds[d].Name})
+			}
```
mittachaitu: I have little concern over here... Let's say that the user is given to pick 4(nDevices) devices from a node to form a pool then I will suggest forming (nDevices/2) raidgroups instead of accumulating on same raidGroup

```go
   // Ignore this check if it already made in caller functions
   // Check if required bd's exist for given node
   if len(bds) < nDevices {
      return errors.New("required no.of(%d) blockdevices doesn't exist on node %s", nDevices, hostName)
   }
   bdIndex := 0
   var rgs []cstorv1.RaidGroup
  for rgIndex :=0; rgIndex<(nDevices/2); rgIndex++ {
          var raids []cstorv1.CStorPoolInstanceBlockDevice
          for d:=0; d<defaultMirrorCount; d++ {
                 raids = append(raids, cstorv1.CStorPoolInstanceBlockDevice{BlockDeviceName: bds[bdIndex].Name})
                 bdIndex++
          }
          rgs = append(rgs, raids)
  }
```
WDYT on above part?


Basically, with the current approach if the user configures 4 devices of size 10G then the size of the pool will be 10G(n redundancy will exist at pool level)
```sh
sparse-01d1417508c6ee62a915bfd0c341c78f        centos-worker-1   /var/openebs/sparse/2-ndm-sparse.img                10737418240      Unclaimed    Active     3h20m
sparse-07a7fbcd6a14cd09e21a539734ca4772        centos-worker-1   /var/openebs/sparse/4-ndm-sparse.img                10737418240      Unclaimed    Active     40s
sparse-5ac9c0a5b87adeecba382f3ec1c093d5        centos-worker-1   /var/openebs/sparse/1-ndm-sparse.img                10737418240      Unclaimed    Active     3h20m
sparse-9903f755f31c265bb190df228f2d3221        centos-worker-1   /var/openebs/sparse/5-ndm-sparse.img                10737418240      Unclaimed    Active     40s
sparse-f62a0627a72f86deba09e73f8f2dc809        centos-worker-1   /var/openebs/sparse/3-ndm-sparse.img                10737418240      Unclaimed    Active     3h20m

## Creating mirror pool with 4 blockdevices
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zpool create pool1 mirror /var/openebs/sparse/2-ndm-sparse.img  /var/openebs/sparse/4-ndm-sparse.img /var/openebs/sparse/1-ndm-sparse.img /var/openebs/sparse/5-ndm-sparse.img

### size of the pool is 10G
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zfs list
NAME                                         USED  AVAIL  REFER  MOUNTPOINT
pool1                                         53K  9.63G   512B  /pool1

### Pool vdev will looks like
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zpool status -v
  pool: pool1
 state: ONLINE
  scan: none requested
config:

	NAME                                      STATE     READ WRITE CKSUM
	pool1                                     ONLINE       0     0     0
	  mirror-0                                ONLINE       0     0     0
	    /var/openebs/sparse/2-ndm-sparse.img  ONLINE       0     0     0
	    /var/openebs/sparse/4-ndm-sparse.img  ONLINE       0     0     0
	    /var/openebs/sparse/1-ndm-sparse.img  ONLINE       0     0     0
	    /var/openebs/sparse/5-ndm-sparse.img  ONLINE       0     0     0
```

With above suggestion it will be create pool of 20G with same configuration but redundancy of data will be only 1 copy than 3 copies
```sh
## Creating stripe pool on top of mirror pool
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zpool create pool1 mirror /var/openebs/sparse/2-ndm-sparse.img  /var/openebs/sparse/4-ndm-sparse.img mirror /var/openebs/sparse/1-ndm-sparse.img /var/openebs/sparse/5-ndm-sparse.img

## size of pool is 20G
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zfs list
NAME                                         USED  AVAIL  REFER  MOUNTPOINT
pool1                                         50K  19.3G   512B  /pool1

## pool vdev will looks like below
root@cstor-disk-cspc-v5jb-7f7d498b47-t8w58:/# zpool status
  pool: pool1
 state: ONLINE
  scan: none requested
config:

	NAME                                      STATE     READ WRITE CKSUM
	pool1                                     ONLINE       0     0     0
	  mirror-0                                ONLINE       0     0     0
	    /var/openebs/sparse/2-ndm-sparse.img  ONLINE       0     0     0
	    /var/openebs/sparse/4-ndm-sparse.img  ONLINE       0     0     0
	  mirror-1                                ONLINE       0     0     0
	    /var/openebs/sparse/1-ndm-sparse.img  ONLINE       0     0     0
	    /var/openebs/sparse/5-ndm-sparse.img  ONLINE       0     0     0

errors: No known data errors
```

mittachaitu (commented):
Provided few comments/suggestions... PTAL

cmd/generate/generate.go
vharsh: I'll make this change as we aren't sure if mayastor functionalities will be included in this CLI.

cmd/generate/generate.go
vharsh: I'll see if a function taking `cmd` as an argument and returning the default minimum number of disks can be returned, good thought.

pkg/generate/cspc.go
vharsh: I have this function to call different generator functions for `cstor`, but it seems pool concept as such isn't there on non-cstor nodes wherein writing Pool YAMLs is complex.

pkg/generate/cspc.go
vharsh: I wonder why golangci-lint doesn't raise such warnings :(

pkg/generate/cspc.go
vharsh: given the context is an empty one, i.e. it'll get stuck sort of forever and not get cancelled, what other `err` can come when `len(pod.Items)` is non zero?

vharsh: Ah, got it, even when the pods won't exits, the `err` will likely be nil, I'll make the necessary changes.

pkg/generate/cspc.go
vharsh: Okay, that shouldn't be <10 lines of code, so I'll do the following for every-case, as it might not be straightforward to figure out if a cluster has this quirk. Found relevant bug @ `https://github.com/kubernetes/kubernetes/issues/54482`

1. Get all nodes by the node-name and then find the hostname and put them in a map[nodeName]HostName.
2. Assuming, `kubectl get nodes` gets me the `metadata.name` which is the nodeName and `metadata.Labels['kubernetes.io/hostname'] will get me the hostName, which I can map and then put them in a comma separated string

Or best, I don't even need a map 🤷🏼

pkg/generate/cspc.go
vharsh: Okay, so you mean, I should cap the number of BDs in a RaidGroup to 2? If a user gives 6 `nDevices`, 3 RaidGroups should be formed in each pool instance spec?

mittachaitu: >  Okay, so you mean, I should cap the number of BDs in a RaidGroup to 2? If a user gives 6 nDevices, 3 RaidGroups should be formed in each pool instance spec?

Yes, this way cli will offer best to user out of given things

vharsh: Got, it will do the same as suggested.

pkg/generate/cspc.go
vharsh: TODO: @vharsh, evaluate the cons of filtering out bad nodes with respect to code, UT & product usage.

cmd/generate/generate.go
vharsh: Changed it to `NewCmdGenerateCstorStoragePoolCluster` to avoid naming convention with earlier name of pool

pkg/generate/cspc.go
vharsh: I've renamed the functions to `CSPC` and `cspc`, they aren't good names, but having a function create a K8s-client object helps me better unit test it.

Background: The CSPC function is a dumb one, it does two things
1. Create a client object
2. Call the `cspc` which does interesting stuff and returns the `err` which is printed to the console.

Benefits: I can easily write the tests for business logic( which is in the `cspc(....)` and `CSPC()` has stuff which don't require tests). This way the `cmd` package has no k8s dependency and `cmd` package tests can also be written as an overkill someday, moreover, someone can import `openebsctl` as a dependency and call some of it's function(won't happen anyday I suppose).

pkg/client/k8s.go
vharsh: I'll put them in a map of string -> pointer to node instead, I wish there was an easy way to do this without resorting to unstructured objects(the java dev inside me misses generics).

pkg/generate/cspc.go
vharsh: [Update] I've done most of the work, I'm done stuck between this nodename & hostname quirk. Usually peeps would talk in terms of nodeName(because that's what they see in `kubectl get nodes`), so it'd make sense I show errors with nodeNames. Aah, this can get tricky, I need a map of `nodeName->hostName` or vice-versa. 😞

vharsh: Wrote similar logic, since I know that the number of BDs in a RaidGroup will be fixed to `2`, i.e. default one, I'm just adding two BDs into the RaidGroup slice at a time.

pkg/generate/testdata_test.go
```diff
@@ -0,0 +1,301 @@
...
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n1"}, {BlockDeviceName: "bd2-n1"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}},
+		{NodeSelector: map[string]string{"kubernetes.io/hostname": "node2"},
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n2"}, {BlockDeviceName: "bd2-n2"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}},
+		{NodeSelector: map[string]string{"kubernetes.io/hostname": "node3"},
+			DataRaidGroups: []cstorv1.RaidGroup{{CStorPoolInstanceBlockDevices: []cstorv1.CStorPoolInstanceBlockDevice{{
+				BlockDeviceName: "bd1-n3"}, {BlockDeviceName: "bd2-n3"}}}}, PoolConfig: cstorv1.PoolConfig{DataRaidGroupType: string(cstorv1.PoolMirrored)}}}}}
+
+var mirrorCSPCFourBDs = cstorv1.CStorPoolCluster{
```
vharsh, line 161: @mittachaitu Does this mirror CSPC on node1 & node2 look good where-in each RaidGroup has 2 blockdevices?

For six BDs, there'd be 3 RaidGroups per instance/Node, right?

pkg/generate/cspc.go
mittachaitu: Yeah, Good to reject sparse-based blockdevices(Good Point :+1: ),

**Just a optional suggestion**

We can replace the above snippet with something like below(it can be improved further):
```go
type poolBDFilter func(v1alpha1.BlockDevice) bool
var (
poolEligibleBDFilters := []poolBDFilter{
  isBlockDeviceActive,
  isBlockDeviceUnclaimed,
  isBlockDeviceUnFormatted,
  isNonSparseBlockDevice,
  isBlockDeviceHasMinCap(cap uint64) func(v1alpha1.BlockDevice) bool {
      return func(bd v1alpha1.BlockDevice) bool {
           return bd.Spec.Capacity.Storage >= minCapacity
      }
  }
}
)

func isBlockDeviceEligibleForPoolCreation(bd v1alpha1.BlockDevice, filters []poolBDFilter) bool {
             for i:=0; i<len(filters); i++ {
                   if !filters[i](bd) {
                       return false
                  }
             }
             return true
}

if isBlockDeviceEligibleForPoolCreation(bd, poolEligibleBDFilters) {
   final = append(final, bd)
}
```

pkg/generate/testdata_test.go
mittachaitu: Yes

pkg/generate/cspc.go
```diff
@@ -0,0 +1,243 @@
...
+	// visible clearly for EKS clusters
+	var hostnames []string
+	for _, node := range nodeList.Items {
+		// I hope it is unlikely for a K8s node to have an empty hostname
+		hostnames = append(hostnames, node.Labels["kubernetes.io/hostname"])
+	}
+	// 2. Fetch BD's from the eligible/valid nodes by hostname labels
+	bds, err := c.GetBDs(nil, "kubernetes.io/hostname in ("+strings.Join(hostnames, ",")+")")
+	if err != nil || len(bds.Items) == 0 {
+		return nil, "", fmt.Errorf("no blockdevices found in nodes with %v hostnames", hostnames)
```
mittachaitu, line 83: Good to log error also

mittachaitu (approved):
LGTM

mittachaitu (commented):
We missed picking same size of blockdevices and filtering sparse blockdevices instead of combining them with non-sparse blockdevices

pkg/generate/cspc.go
vharsh: Returned errors are printed, the chain is a bit complex to the point it's not easy to figure these things out(which is not a good practice IMO) 😰 .

cspc.go:83 -> cspc.go:49 -> cmd/generate/generate.go:54 -> `util.CheckError is called with util.Fatal` and then this error is printed to stderr from pkg/util/util.go:74

vharsh:
> We missed picking same size of blockdevices and filtering sparse blockdevices instead of combining them with non-sparse blockdevices

@mittachaitu I'm planning to handle those things in upcoming PR as this PR has become a bit big

pkg/generate/cspc.go
vharsh: this error will get logged eventually as this line of errors aren't swallowed by the line of callers

pkg/generate/cspc.go
vharsh: implemented similar logic to generate, RaidZ & RaidZ2 pools, I'll now have to do something to fix the capacity of the BDs which get selected in the pool assuming two cases
1. I have an infinite supply of BDs of different sizes
2. I have a finite supply of BDs and I should backtrack and re-select the most optimal ones which also minimize disk wastage

PR #126, opened 2021-10-09
chore(tools): Add gitpod development support

### LINKED ISSUE #125: Gitpodify openebs/openebsctl for easier contribution
Setting up environment like installing various dependencies, having right version of tools might discourage potential contributors. Having this repo gitpodified would set up dev env with one click.

One of the repository is already gitpodified.

https://github.com/openebs/cstor-operators/blob/develop/.gitpod.yml

princerachit:
Fixes https://github.com/openebs/openebsctl/issues/125

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,14 @@
+tasks:
+  - init: go mod vendor
+github:
+  prebuilds:
+    master: true
+    branches: true
+    pullRequests: true
+    pullRequestsFromForks: true
+    addCheck: true
+
+vscode:
+  extensions:
+    - golang.go
+    - ms-azuretools.vscode-docker


.gitpod.yml
```diff
@@ -0,0 +1,14 @@
+tasks:
+  - init: make all
+github:
+  prebuilds:
+    master: true
```
vharsh, line 5: Should this be `develop` as there's no `master` branch

.gitpod.yml
```diff
@@ -0,0 +1,14 @@
+tasks:
+  - init: make all
```
vharsh: This can be changed to `go mod vendor` command, I just observed, Makefile might also need some cleaning

.gitpod.yml
princerachit: https://www.gitpod.io/docs/prebuilds#configure-the-github-app

master here refers to the default branch of the repo not the actual name `master`

PR #127, opened 2021-10-16
[refractor] break k8s monolith into smaller components

Abhishek-kumar09:
Broken k8s client into components:
- bd
- jiva
- cstor
- k8s cores


pkg/client/bd_client.go
```diff
@@ -0,0 +1,92 @@
+/*
```
vharsh: Please rename this file to `bd.go`

Reason: the package name already has the word `client`, the `client` and `bd_client` repeat by a bit, which is sort of unnecessary.

pkg/client/volume_client.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2020-2021 The OpenEBS Authors
```
vharsh: this file contents can be moved to the now `k8s_core.go` as the volumes & the volumeClaim resources are a part of the `CoreV1` API-group.

pkg/client/k8s_core.go
```diff
@@ -0,0 +1,85 @@
+/*
```
vharsh: This file can be renamed to `k8s.go`.

pkg/client/volume_client.go
vharsh: I think `k8s_client.go`, `k8s_core.go`, `volume_client.go` can be merged into a single file.

vharsh (approved):

vharsh (approved):
lgtm

Issue #59, opened 2021-07-17
Enhance describe volume and pvc to display the pod(s) consuming the PVC


kmova:
The current command shows the volume details and the node to which a PV is attached. 

In the case of RWX volumes, there can be multiple nodes to be shown. Extending this thought, it will be nice to show the pods that are consuming the PV.

daniel-shuy:
I've created a PR (#129) to show the Pod(s) mounting the PVC in `describe pvc`.

I didn't implement it for `describe volume` as I am not sure how to get a Pod's PV.

PR #129, opened 2021-10-21
feat(pvc describe): Add mount pods

daniel-shuy:
Partially implements #59

Enhances `describe pvc` to display all pods that are mounting the PVC, eg.
```
MOUNTED BY       : pod-a pod-b 
```

Because the mount pod information is not added to the [PersistentVolumeClaim ](https://github.com/kubernetes/api/blob/v0.20.2/core/v1/types.go#L433) object in [k8s.io/api](https://github.com/kubernetes/api), I had to loop through all Pods, similar to what `kubectl describe pvc` is doing (see kubernetes/kubernetes#65837).

I only implemented it for `describe pvc`, but not `describe volume`, as I am not sure how to get a Pod's PV.

### DIFF
--- a/pkg/persistentvolumeclaim/persistentvolumeclaim.go
+++ b/pkg/persistentvolumeclaim/persistentvolumeclaim.go
@@ -17,6 +17,8 @@ limitations under the License.
 package persistentvolumeclaim
 
 import (
+	"sort"
+
 	"github.com/openebs/openebsctl/pkg/client"
 	"github.com/openebs/openebsctl/pkg/util"
 	"github.com/pkg/errors"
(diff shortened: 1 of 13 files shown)


pkg/persistentvolumeclaim/persistentvolumeclaim.go
```diff
@@ -70,12 +81,51 @@ func Describe(pvcs []string, namespace string, openebsNs string) error {
...
+func PodsToString(pods []corev1.Pod) string {
+	if len(pods) == 0 {
+		return "<none>"
+	}
+	str := ""
+	for _, pod := range pods {
+		str += pod.Name + " "
+	}
+	return str
+}
```
daniel-shuy, line 131: I chose to display it as a space delimited string to be in line with `checks.AccessModeToString()`:
https://github.com/openebs/openebsctl/blob/ac15f30b3644c2eda0bf9da25bf7d239c38876ea/pkg/util/checks.go#L49-L56

Let me know if a different format is desired (eg. comma/newline delimited string).

vharsh:
@daniel-shuy You might have missed to commit the changes from `pkg/persistentvolumeclaim/generic_test.go`, that's why the builds seem to have failed.

daniel-shuy:
@vharsh oops, you're right, I've committed it, thanks!

pkg/persistentvolumeclaim/persistentvolumeclaim.go
vharsh: Space delimited pod name strings are awesome, thanks 👍🏼

vharsh (approved):
lgtm

PR #130, opened 2021-10-23
(feat) Add support for log and log streams for upgrade jobs (jiva)

### LINKED ISSUE #128: Add feature to view the jiva volume upgrade status
As per https://github.com/openebs/openebsctl/pull/115 the feature has been implemented to schedule the upgrade job to upgrade the Jiva volume data plane components to match the version of the jiva control plane. @Abhishek-kumar09 has suggested for a way to add a watch functionality to either 
- see the `upgrade status` as the job runs to successful completion or failure 
- or to tail the logs while running an upgrade

Both of the above are great ideas to simplify the UX for launching upgrade jobs for data plane, this issue will house the discussion for the same.

Continued from the discussions at #115

Abhishek-kumar09:
Partially Fixes #128 

The logs are streaming from the upgrade job running in the cluster.
- [x] Add subcommand for upgrade status
- [x] Instantiate log support: Outputs all the logs from the pods and containers running in a job and return immediately
- [x] --wait flag for starting a log stream that watches the logs and remains active in the terminal until ctrl+C is pressed or container has stopped.
- [x] No need to provide job name, automatically determines the jobs running for the upgrade purpose and outputs the logs from all the pods and all the contianers registered under the job
- Active log Streaming is only available  for one pod in a job and optimized by 2 sec sleep to save CPU resources.

Instantaneos logs:
```console
$ ./k upgrade status --cas-type jiva --openebs-namespace openebs
***************************************
Job Name:  jiva-upgrade-lt6kq
From Pod: jiva-upgrade-lt6kq-hr6cp
F1023 15:24:16.930123       1 jiva_volume.go:52] Invalid from version 2.12.2 or to version 2.10.0

***************************************
```
Start Log Stream
```console
$ ./k upgrade status --cas-type jiva --openebs-namespace openebs --wait
Waiting for the Pod Logs...
F1023 15:24:16.930123       1 jiva_volume.go:52] Invalid from version 2.12.2 or to version 2.10.0
...
```
Next Task: 
- Start event log-stream from jobs (Maybe I will do it in separate PR)


cmd/upgrade/upgrade.go
```diff
@@ -53,18 +53,19 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
 				fmt.Println("To upgrade other cas-types follow: https://github.com/openebs/upgrade#upgrading-openebs-reources")
 			}
 		},
 	}
+	cmd.AddCommand(NewCmdUpgradeStatus())
 	cmd.SetUsageTemplate(upgradeCmdHelp)
+	cmd.PersistentFlags().StringVarP(&upgrade.OpenebsNs, "openebs-namespace", "", "", "provide openebs-namespace")
 	cmd.PersistentFlags().StringVarP(&upgrade.CasType, "cas-type", "", "", "the cas-type filter option for fetching resources")
 	cmd.PersistentFlags().StringVarP(&upgrade.ToVersion, "to-version", "", "", "the version to which the resources need to be upgraded")
 	cmd.PersistentFlags().StringVarP(&upgrade.File, "file", "f", "", "provide path/url to the menifest file")
```
Abhinandan-Purkait: s/menifest/manifest

cmd/upgrade/status.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewCmdUpgradeStatus() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "status",
+		Aliases: []string{"Status"},
+		Short:   "Display Upgrade-status for a running upgrade-job",
+		Run: func(cmd *cobra.Command, args []string) {
+			if upgrade.CasType != util.JivaCasType {
```
vharsh: Considering the fact that as of now, non-jiva upgrades aren't supported, please call the Jiva upgrade job status even when `upgade.CasType` is `""`.

pkg/upgrade/jiva.go
```diff
@@ -83,20 +83,24 @@ func InstantiateJivaUpgrade(openebsNs string) {
 			return
 		}
 
+		//  GetJobStatus()
```
Abhinandan-Purkait: commented code? Either remove this or add a TODO if its something you want to revisit later

cmd/upgrade/status.go
```diff
@@ -0,0 +1,44 @@
...
+				fmt.Println("Only jiva upgrade-status are supported!")
+				os.Exit(1)
+			}
+			status.GetJobStatus()
+		},
+	}
+	cmd.PersistentFlags().BoolVar(&status.WaitFlag, "wait", false, "Wait for the logs stream")
```
vharsh: @kmova This status sub-command's functionality seems a bit analogous to `get`(get me a list of interesting things) and `describe`(describe one of the things in detail) when the `--wait` flag is present, is the current CLI UX okay, or should we change some UX?

pkg/client/k8s.go
```diff
@@ -470,9 +473,9 @@ func (k K8sClient) GetBatchJob(name string, namespace string) (*batchV1.Job, err
 	return job, nil
 }
 
-// GetBatchJobs returns all the batch jobs running in all-namespaces
-func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
-	list, err := k.K8sCS.BatchV1().Jobs("").List(context.Background(), metav1.ListOptions{})
+// GetBatchJobs returns batch jobs running in the namespacec with the label
```
Abhinandan-Purkait: s/namespacec/namespace

pkg/upgrade/jiva.go
```diff
@@ -83,20 +83,24 @@ func InstantiateJivaUpgrade(openebsNs string) {
...
+		// fmt.Println(k.GetPodLogs(pods.Items[0], openebsNs))
+		// k.StartPodLogsStream(pods.Items[0], openebsNs)
+
 		ToVersion = pods.Items[0].Labels["openebs.io/version"]
 	}
 
 	// assign namespace
-	if openebsNs == "" {
+	if OpenebsNs == "" {
 		fmt.Println(`No Namespace Provided, using "default" as a namespace`)
```
vharsh: Please get the openebs-namespace via utility methods by using the namespace of JIVA CSI drivers.

pkg/upgrade/jiva.go
```diff
@@ -250,7 +254,8 @@ func getContainerArguments(cfg *jivaUpdateConfig) []string {
 }
 
 func checkIfJobIsAlreadyRunning(k *client.K8sClient, cfg *jivaUpdateConfig) (bool, error) {
-	jobs, err := k.GetBatchJobs()
+	// get all jobs in all namespaces
+	jobs, err := k.GetBatchJobs("", "")
```
vharsh: Filtering by the jiva upgrade job labels would be a good idea here.

pkg/upgrade/status/status.go
```diff
@@ -0,0 +1,105 @@
...
+	// get pods created by the job
+	podList, err := k.GetPods(fmt.Sprintf("job-name=%s", name), "", namespace)
+	if err != nil {
+		fmt.Println("error getting pods of job", name, ": err", err)
+		return
+	}
+
+	// range over pods to get all the logs
+	for _, pod := range podList.Items {
+		fmt.Println("From Pod:", pod.Name)
```
Abhinandan-Purkait: Can we color this text so that its more readable as the logs would overload the screen, and it would be easier to find it that way.
ref:- https://github.com/openebs/openebsctl/blob/70eeb5ab6fac38acb0d5cf76bf1f02cb49f39b42/pkg/util/util.go#L56

pkg/client/k8s.go
```diff
@@ -484,3 +487,60 @@ func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
 func (k K8sClient) DeleteBatchJob(name string, namespace string) error {
 	return k.K8sCS.BatchV1().Jobs(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
 }
+
+// GetPodLogs returns logs from the containers running in the pod
+func (k K8sClient) GetPodLogs(pod corev1.Pod, ns string) string {
```
vharsh: Just sending out the pod name might be a good idea.

pkg/upgrade/status/status.go
```diff
@@ -0,0 +1,105 @@
+/*
```
vharsh: This file can be named as `jiva.go` as this is only related to Jiva volumes.

pkg/upgrade/status/status.go
```diff
@@ -0,0 +1,105 @@
...
+	// No jobs found
+	if len(joblist.Items) == 0 {
+		fmt.Printf("No upgrade-jobs Found in %s namespace", upgrade.OpenebsNs)
+	}
+
+	if WaitFlag {
+		startLogStream(k, joblist)
+	}
+
+	for _, job := range joblist.Items {
```
Abhinandan-Purkait: Trying to understand the flow here, incase the wait flag is true, the log stream of first job shows up, then again this loop here shows the instantaneous log for each job. May not be a good UX. cc @vharsh

pkg/upgrade/status/status.go
```diff
@@ -0,0 +1,105 @@
...
+		fmt.Println(logs)
+	}
+
+	if len(podList.Items) == 0 {
+		fmt.Println("No pods are running for this job")
+	}
+}
+
+// startLogStream starts opens log stream for a pod
+func startLogStream(k *client.K8sClient, jobList *batchV1.JobList) {
```
Abhinandan-Purkait: You can take a `jobName` in the parameter instead of taking a `jobList` as either way it gets the first `jobName` and shows logs for that. Function signature is a bit unclear.

pkg/upgrade/status/status.go
```diff
@@ -0,0 +1,105 @@
...
+		fmt.Println("Error getting jiva-upgrade jobs:", err)
+		return
+	}
+
+	// No jobs found
+	if len(joblist.Items) == 0 {
+		fmt.Printf("No upgrade-jobs Found in %s namespace", upgrade.OpenebsNs)
+	}
+
+	if WaitFlag {
```
vharsh: How will this handle multiple job logs? That'd require some deep thinking.

pkg/upgrade/status/status.go
vharsh: Right, either all jobs should be listed or a single job's logs should be printed.

pkg/upgrade/status/status.go
Abhishek-kumar09: Does go-channels is the expectation here? 
Concurrency is the thing that is most suitable in the first look.

pkg/upgrade/jiva.go
Abhishek-kumar09: Doing it in a seperate PR. This requires updates in job creation and handling.

pkg/upgrade/jiva.go
Abhishek-kumar09: Doing it in seperate PR

pkg/upgrade/status/status.go
Abhishek-kumar09: Added return statement for handling it :)

pkg/upgrade/status/status.go
Abhishek-kumar09: Done.

pkg/upgrade/jiva.go
Abhishek-kumar09: Left by mistake, corrected

pkg/client/k8s.go
```diff
@@ -484,3 +487,58 @@ func (k K8sClient) GetBatchJobs() (*batchV1.JobList, error) {
...
+			break
+		}
+
+		// Err fetching logs
+		if err != nil {
+			log.Fatal("stopping pod logs stream, err:", err)
+		}
+
+		message := string(buf[:numBytes])
+		fmt.Print(util.ColorText(message, util.Blue))
```
Abhishek-kumar09, line 539: Using Blue color now for log messages, another alternative could be green too.

pkg/upgrade/status/status.go
vharsh: No, I was expecting we'd only be streaming one job's log at a time, I don't think it'd be possible to start a stream of multiple upgrade jobs without mixing them up.

cmd/upgrade/status.go
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+package upgrade
+
+import (
+	"github.com/openebs/openebsctl/pkg/upgrade/status"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdUpgradeStatus() *cobra.Command {
```
vharsh, line 24: Can this command be removed in favour of `--wait` flag in the upgrade sub-command itself?
This will remove a requirement for another command, and will sort of match the recommended way suggested by [cobra CLI's UX and a lot of other *nix tool UX](https://github.com/spf13/cobra#concepts).

cmd/upgrade/status.go
vharsh: Preferring a --wait flag instead of a separate command for the same.

cmd/upgrade/status.go
Abhishek-kumar09: Are you asking to remove the command completely or just the --wait flag for the command and shift it to the job creation time?

Abhishek-kumar09: There are two portions being one that shows logs and other that waits for the continuous logs.

pkg/client/k8s.go
```diff
@@ -482,9 +485,60 @@ func (k K8sClient) DeleteBatchJob(name string, namespace string) error {
...
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		log.Fatal("err converting log buffer", err)
+	}
+
+	return buf.String()
+}
+
+// StartPodLogsStream starts stream of logs for the given pod
+func (k K8sClient) StartPodLogsStream(podName string, ns string) {
```
Abhishek-kumar09, line 507: Left this function to be used for streaming logs in some upcoming PR. It is not used rn.

Abhishek-kumar09 (commented):
Removed --wait flag and it's just the status subcommand with upgrade command that shows container logs of any number of running pods inside a batch-job

pkg/upgrade/status/jiva.go
```diff
@@ -0,0 +1,79 @@
...
+		fmt.Println("Job Name: ", job.Name)
+		getPodLogs(k, job.Name, namespace)
+	}
+	fmt.Println("***************************************")
+}
+
+// Get all the logs from the pods associated with a job
+func getPodLogs(k *client.K8sClient, name string, namespace string) {
+	// get pods created by the job
+	podList, err := k.GetPods(fmt.Sprintf("job-name=%s", name), "", namespace)
```
vharsh, line 54: 👍🏼

vharsh (approved):

PR #133, opened 2021-10-30
usablility: auto-determine openebs namespace for job creation

Abhishek-kumar09:
Autodetermine openebs namespace while creating upgrade jobs

### DIFF
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -48,15 +48,16 @@ const (
 
 // NewCmdVolumeUpgrade to upgrade volumes and interfaces
 func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
+	upgradeOpts := upgrade.UpgradeOpts{}
 	cmd := &cobra.Command{
 		Use:     "upgrade",
 		Short:   "Upgrade Volumes, storage engines, and interfaces in openebs application",
 		Aliases: []string{"update"},
 		Run: func(cmd *cobra.Command, args []string) {
-			if !util.IsValidCasType(upgrade.CasType) {
-				fmt.Fprintf(os.Stderr, "cas-type %s not supported\n", upgrade.CasType)
-			} else if upgrade.CasType == util.JivaCasType {
-				upgrade.InstantiateJivaUpgrade()
+			if !util.IsValidCasType(upgradeOpts.CasType) {
+				fmt.Fprintf(os.Stderr, "cas-type %s not supported\n", upgradeOpts.CasType)
+			} else if upgradeOpts.CasType == util.JivaCasType {
+				upgrade.InstantiateJivaUpgrade(upgradeOpts)
 			} else {
 				fmt.Println("Only Jiva upgrades are available at this point")
 				fmt.Println("To upgrade other cas-types follow: https://github.com/openebs/upgrade#upgrading-openebs-reources")
@@ -65,11 +66,9 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
 	}
 	cmd.SetUsageTemplate(upgradeCmdHelp)
 	// using openebs as default value to namespace
-	cmd.PersistentFlags().StringVarP(&upgrade.OpenebsNs, "openebs-namespace", "", "openebs", "provide openebs-namespace")
-	cmd.PersistentFlags().StringVarP(&upgrade.CasType, "cas-type", "", "", "the cas-type filter option for fetching resources")
-	cmd.PersistentFlags().StringVarP(&upgrade.ToVersion, "to-version", "", "", "the version to which the resources need to be upgraded")
-	cmd.PersistentFlags().StringVarP(&upgrade.File, "file", "f", "", "provide path/url to the menifest file")
-	cmd.PersistentFlags().StringVarP(&upgrade.ImagePrefix, "image-prefix", "", "", "provide image prefix for the volume deployments")
-	cmd.PersistentFlags().StringVarP(&upgrade.ImageTag, "image-tag", "", "", "provide custom image tag for the volume deployments")
+	cmd.PersistentFlags().StringVarP(&upgradeOpts.CasType, "cas-type", "", "", "the cas-type filter option for fetching resources")
+	cmd.PersistentFlags().StringVarP(&upgradeOpts.ToVersion, "to-version", "", "", "the version to which the resources need to be upgraded")
+	cmd.PersistentFlags().StringVarP(&upgradeOpts.ImagePrefix, "image-prefix", "", "", "provide image prefix for the volume deployments")
+	cmd.PersistentFlags().StringVarP(&upgradeOpts.ImageTag, "image-tag", "", "", "provide custom image tag for the volume deployments")
 	return cmd
 }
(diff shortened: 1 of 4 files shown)


cmd/upgrade/upgrade.go
```diff
@@ -53,18 +53,19 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
+				upgrade.InstantiateJivaUpgrade()
 			} else {
 				fmt.Println("Only Jiva upgrades are available at this point")
 				fmt.Println("To upgrade other cas-types follow: https://github.com/openebs/upgrade#upgrading-openebs-reources")
 			}
 		},
 	}
 	cmd.SetUsageTemplate(upgradeCmdHelp)
+	// using openebs as default value to namespace
+	cmd.PersistentFlags().StringVarP(&upgrade.OpenebsNs, "openebs-namespace", "", "openebs", "provide openebs-namespace")
```
Abhishek-kumar09: I have used "openebs" as a default namespace, since all the docs are written with this consideration. The more chances are openebs will be the default namespace for all openebs components.

pkg/upgrade/jiva.go
```diff
@@ -250,7 +254,7 @@ func getContainerArguments(cfg *jivaUpdateConfig) []string {
 }
 
 func checkIfJobIsAlreadyRunning(k *client.K8sClient, cfg *jivaUpdateConfig) (bool, error) {
-	jobs, err := k.GetBatchJobs()
+	jobs, err := k.GetBatchJobs("", "cas-type=jiva,name=jiva-upgrade")
```
Abhishek-kumar09, line 203: Applied: https://github.com/openebs/openebsctl/pull/130#discussion_r735282615

Considering the fact that we have filtered out the pods with the labels,
it comes with the problem that if users are upgrading jobs with their own yaml files or following default doc, it won't be able to track those running pods, since they won't be having any labels as such.

pkg/upgrade/jiva.go
```diff
@@ -50,9 +50,19 @@ type jobInfo struct {
...
 // Jiva Data-plane Upgrade Job instantiator
-func InstantiateJivaUpgrade(openebsNs string) {
+func InstantiateJivaUpgrade() {
 	k := client.NewK8sClient()
 
+	// assign namespace
+	if OpenebsNs == "openebs" {
+		// namespace equals to default val -> not provided by CLI flags
+		// auto-determine jiva namespace
+		ns, err := k.GetOpenEBSNamespace(util.ZFSCasType)
```
vharsh:
```suggestion
		ns, err := k.GetOpenEBSNamespace(util.JivaCasType)
```

cmd/upgrade/upgrade.go
vharsh: set it to `""`, this way one can get to know if `openebs-namespace` has actually been provided via flags or the default value has been used

pkg/client/k8s.go
```diff
@@ -190,7 +190,12 @@ func (k K8sClient) GetOpenEBSNamespace(casType string) (string, error) {
 	if err != nil || len(pods.Items) == 0 {
 		return "", errors.New("unable to determine openebs namespace")
 	}
-	return pods.Items[0].Namespace, nil
+	ns := pods.Items[0].Namespace
```
vharsh: Why is this change required? 

[K8s] Usually `default` namespace is used while creating resources if the resource namespace isn't specified by the client.

<br/>
<br/>
<br/>
<br/>

Pods(are namespaced resources) will have a a non-empty, namespace. Here's an example of a pod in default namespace. 

```bash
➜  $ kubectl get po csi-6447f6bfcf-mch72 -oyaml | grep namespace
  namespace: default
```

pkg/upgrade/jiva.go
```diff
@@ -50,9 +50,19 @@ type jobInfo struct {
 }
 
 // Jiva Data-plane Upgrade Job instantiator
-func InstantiateJivaUpgrade(openebsNs string) {
+func InstantiateJivaUpgrade() {
 	k := client.NewK8sClient()
 
+	// assign namespace
+	if OpenebsNs == "openebs" {
```
vharsh: as discussed previously, this will have to check for `""`

vharsh (commented):
Great refactors 🎉🔧, minor comments.

pkg/upgrade/jiva.go
vharsh: Fair point, I'll confirm with the team once.

cmd/upgrade/upgrade.go
Abhishek-kumar09: So this is done to shorten the code logic. Default value is "openebs" as all the openebs docs have this as their default namespace value.
If user is not putting any value then "openebs" is used.
The other way, we have to write additional if statement to check if the namespace is provided or not and assign it as "openebs" if not. This is just to get rid of that one if statement and use the default values feature provided by the framework

pkg/client/k8s.go
Abhishek-kumar09: Okay this is due to fact that the function return namespace of the case types provided to it. 
Ideally it should return error if the cas-types are not present in the cluster doesn't have any namespace.

e.g., I don't have any zfs volumes present in the cluster, but when I call this function to get the namespace of the zfs volume it returns an empty string as the namespace. If it would provided error that "resource is not deployed in the cluster  hence doesn't have any namespace associated with it", then it would have been more desirable thing.

vharsh: > Ideally it should return error if the cas-types are not present in the cluster

Correct, it will return `"", err` if the cas-type is absent in a cluster, i.e. if the ZFS CSI provisioner is not present, it'd likely return a `non nil error` or at least the len(pods.Items) will be zero(check the conditional statement in line#190).

Have you seen it otherwise here(in that case there'd be a bug in the code which we'd have to fix).

But if `pods.Items[0]` is actually a pod, we'd get a non `""` namespace for sure.

vharsh: One alternative way to think of it might be to change the return type of the function to `(*string, error)`, i.e. getting a different zero-value. In Golang, it's a bit common to not use the returned value(assume it to be a zero-value) if the error is non-nil.

cmd/upgrade/upgrade.go
vharsh: that's right, in the operator YAMLs of many cas-types the namespace has been hardcoded to `"openebs"` but in other cases, when installation is done via helm(which is another interesting project), it's really easy to override it e.g. `helm install ABCD --namespace=random`.

pkg/upgrade/jiva.go
```diff
@@ -50,17 +46,14 @@ type jobInfo struct {
 }
 
 // Jiva Data-plane Upgrade Job instantiator
-func InstantiateJivaUpgrade(openebsNs string) {
+func InstantiateJivaUpgrade(upgradeOpts UpgradeOpts) {
```
vharsh, line 49:
```suggestion
func InstantiateJivaUpgrade(opts UpgradeOpts) {
```

the variable name and the variable types stutter a bit, this might be a slightly better name

pkg/upgrade/jiva.go
```diff
@@ -50,17 +46,14 @@ type jobInfo struct {
...
-	// If manifest Files is provided, apply the file to create a new upgrade-job
-	if File != "" {
-		yamlFile, err := yamlToJobSpec(File)
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "Error in Job: %s", err)
-		}
-		k.CreateBatchJob(yamlFile, yamlFile.Namespace)
-		return
+	// auto-determine jiva namespace
+	ns, err := k.GetOpenEBSNamespace(util.ZFSCasType)
```
vharsh:
```suggestion
	ns, err := k.GetOpenEBSNamespace(util.JivaCasType)
```

Abhishek-kumar09: isss!! left while testing again

vharsh (approved):

PR #134, opened 2021-11-10
[upgrade job] CSPC Pools

Abhishek-kumar09:
```console
$ ./k upgrade --cas-type cstor

Fetching CSPC control plane and Data Plane Version
Current Version: 2.12.0
Desired Version: 3.0.0
Previous job failed.
Reason:  BackoffLimitExceeded
Creating a new Job with name: cstor-cspc-upgrade-vfn87
Creating Dry-run job...
metadata:
  creationTimestamp: null
  generateName: cstor-cspc-upgrade-
  labels:
    cas-type: cstor
    name: cstor-cspc-upgrade
  namespace: openebs
spec:
  backoffLimit: 4
  template:
    metadata:
      creationTimestamp: null
    spec:
      containers:
      - args:
        - cstor-cspc
        - --from-version=2.12.0
        - --to-version=3.0.0
        - --v=4
        - cstor-storage
        env:
        - name: OPENEBS_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        image: openebs/upgrade:3.0.0
        imagePullPolicy: IfNotPresent
        name: upgrade-cstor-cspc-go
        resources: {}
      restartPolicy: OnFailure
      serviceAccountName: openebs-maya-operator
status: {}

Continue?: y
Creating a batch job...
Job Created successfully:  
```


Abhishek-kumar09:
@vharsh up for review.

cmd/upgrade/upgrade.go
```diff
@@ -53,13 +52,22 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
+			case util.JivaCasType:
+				{
+					upgrade.InstantiateJivaUpgrade(upgradeOpts)
+					break
+				}
+			case util.CstorCasType:
+				{
+					upgrade.InstantiateCspcUpgrade(upgradeOpts)
+					break
+				}
```
Abhishek-kumar09: Had one doubt, since CSPC and Cstor jobs will be handled separately, filtering on the basis of cas-type may not work here. Any other approach that you would recommend to follow.

vharsh: Okay, having different labels for both type of cstor jobs might help(i.e. a label key besides the `castype` one).

A Cstor volume upgrade can be thought of as a combination of a Pool upgrade and a Volume upgrade, so there can be upgrade jobs for CStor which have the regular castype label to tell them apart from jiva upgrade jobs and another label to tell each `CStor related object` 's upgrade job.

cmd/upgrade/upgrade.go
```diff
@@ -53,13 +52,22 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
-			if !util.IsValidCasType(upgradeOpts.CasType) {
-				fmt.Fprintf(os.Stderr, "cas-type %s not supported\n", upgradeOpts.CasType)
-			} else if upgradeOpts.CasType == util.JivaCasType {
-				upgrade.InstantiateJivaUpgrade(upgradeOpts)
-			} else {
-				fmt.Println("Only Jiva upgrades are available at this point")
-				fmt.Println("To upgrade other cas-types follow: https://github.com/openebs/upgrade#upgrading-openebs-reources")
+			switch upgradeOpts.CasType {
+			case util.JivaCasType:
+				{
```
vharsh: braces can be avoided for the individual case statements

Ref: https://yourbasic.org/golang/switch-statement/
Ref: https://tour.golang.org/flowcontrol/9

cmd/upgrade/upgrade.go
```diff
@@ -53,13 +52,22 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
+					upgrade.InstantiateJivaUpgrade(upgradeOpts)
+					break
```
vharsh: No `break` required in Golang, one needs to explicitly put `fallthrough` keyword to get behaviour similar to other programming languages though.

Ref: https://tour.golang.org/flowcontrol/9
Ref: https://golangbot.com/switch/

cmd/upgrade/upgrade.go
Abhishek-kumar09: Got it.
I was worried with the command to fire to upgrade cspc pools.
current command is:
```
./k upgrade --cas-type cstor
```

cmd/upgrade/upgrade.go
Abhishek-kumar09: Oh Thanks, that's a good thing being practiced by golang

cmd/upgrade/upgrade.go
Abhishek-kumar09: Ahh super clean

cmd/upgrade/upgrade.go
```diff
@@ -53,13 +52,16 @@ func NewCmdVolumeUpgrade(rootCmd *cobra.Command) *cobra.Command {
...
+			switch upgradeOpts.CasType {
+			case util.JivaCasType:
 				upgrade.InstantiateJivaUpgrade(upgradeOpts)
-			} else {
-				fmt.Println("Only Jiva upgrades are available at this point")
-				fmt.Println("To upgrade other cas-types follow: https://github.com/openebs/upgrade#upgrading-openebs-reources")
+			case util.CstorCasType:
+				upgrade.InstantiateCspcUpgrade(upgradeOpts)
+			default:
+				{
```
vharsh: no brace required in `default` too

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+		k.Ns = "openebs"
+	}
+
+	poolNames := getCSPCPoolNames(k)
+	cfg := UpgradeJobCfg{
+		fromVersion:        "",
+		toVersion:          "",
+		namespace:          k.Ns,
+		resources:          poolNames,
+		serviceAccountName: "openebs-maya-operator",
```
vharsh: this won't be so straightforward again, I might have missed it for Jiva(that might vary too depending on installation method)
Ref: the name might depend on the installation methods -> https://github.com/openebs/charts/blob/gh-pages/microk8s-cstor-operator.yaml#L17

cmd/upgrade/upgrade.go
Abhishek-kumar09: Sorry, left by mistake.

Abhishek-kumar09:
This can  also be added to the project board.

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// auto-determine cstor namespace
+	var err error
+	k.Ns, err = k.GetOpenEBSNamespace(util.CstorCasType)
+	if err != nil {
+		fmt.Println(`Error determining cstor namespace! using "openebs" as namespace`)
+		k.Ns = "openebs"
+	}
+
+	poolNames := getCSPCPoolNames(k)
```
vharsh: Currently the SC's `param.cstorPoolName` is being fetched and clubbed, you can just do `Get` of CSPC directly, this function is not required.

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+		if options.ToVersion == "" {
+			cfg.toVersion = cspc.VersionDetails.Desired // Assigning to-version
+			if cfg.toVersion == "" {
+				continue
+			}
+		} else {
+			cfg.toVersion = options.ToVersion // use cli flag instead
+		}
+		fmt.Println("Desired Version:", cfg.toVersion)
+		break
```
vharsh: This `break` will ensure that this loop runs only once, is this intended by design to find the `toVersion` from CSPC? I think this won't be required after the above change(taking desired version from CSPC operator)

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			fmt.Println("Error detecting version of Cstor Pool Cluster")
+			continue
+		}
+
+		fmt.Println("Fetching CSPC control plane and Data Plane Version")
+		cfg.fromVersion = cspc.VersionDetails.Status.Current // Assigning from Version
+		fmt.Println("Current Version:", cfg.fromVersion)
+		if options.ToVersion == "" {
+			cfg.toVersion = cspc.VersionDetails.Desired // Assigning to-version
```
vharsh: This should be picked up from CSPC operator's `openebs.io/version` label's value, preferably only once outside of the loop.

pkg/upgrade/cstor.go
vharsh: this needs to be dynamically figured out via the CSPC operator's SA name.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,182 @@
...
+func inspectRunningUpgradeJobs(k *client.K8sClient, cfg *UpgradeJobCfg) (bool, error) {
+	jobs, err := k.GetBatchJobs("", "")
+	if err != nil {
+		return false, err
+	}
+
+	// runningJob holds the information about the jobs that are in use by the PV
+	// that has an upgrade-job progress(any status) already going
+	var runningJob *batchV1.Job
+	func() {
```
vharsh, line 65: What is the use of this anon function?

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		} else {
+			cfg.toVersion = options.ToVersion // use cli flag instead
+		}
+		fmt.Println("Desired Version:", cfg.toVersion)
+		break
+	}
+
+	// Check if a job is running with underlying PV
+	res, err := inspectRunningUpgradeJobs(k, &cfg)
```
vharsh: a boolean `res` and an `error`, don't seem to qualify as a good function return type, how about just returning an `error` and deciding things based on that alone?

vharsh: Can this be reduced to a single return value function(just an error) as that is a bit cleaner, moreover there aren't different actions being done for `non-nil-error and false` or `nil error but true`. It's like this method is getting a lot of info but not making any use of it, so it might just be simple to return an `err`.

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+	// If error or upgrade job is already running return
+	if err != nil || res {
+		log.Fatal("An upgrade job is already running with the underlying volume!")
+	}
+
+	// Create upgrade job
+	k.CreateBatchJob(buildCspcbatchJob(&cfg), k.Ns)
+}
+
+func getCSPCPoolNames(k *client.K8sClient) []string {
```
vharsh: this can be deleted as suggested earlier

pkg/upgrade/cstor.go
```diff
@@ -0,0 +1,161 @@
...
+
+// buildCspcbatchJob returns CSPC Job to be build
+func buildCspcbatchJob(cfg *UpgradeJobCfg) *batchV1.Job {
+	return NewJob().
+		WithGeneratedName("cstor-cspc-upgrade").
+		WithLabel(map[string]string{"name": "cstor-cspc-upgrade", "cas-type": "cstor"}). // sets labels for job discovery
+		WithNamespace(cfg.namespace).
+		WithBackOffLimit(cfg.backOffLimit).
+		WithPodTemplateSpec(
+			func() *core.PodTemplateSpec {
```
vharsh, line 89: this is a smart-ish use of closures but it does seem to pop eyeballs at first(to someone not familiar with them), but it's great 👍🏼

pkg/upgrade/cstor.go
vharsh: @Abhishek-kumar09 The CSPC operator will have this label, `openebs.io/component-name=cspc-operator` and the SA name can be found at `deploy.spec.template.spec.serviceAccountName`.

pkg/upgrade/upgrade.go
```diff
@@ -0,0 +1,182 @@
...
+	namespace          string
+	resources          []string
+	backOffLimit       int32
+	serviceAccountName string
+	logLevel           int32
+	additionalArgs     []string
+}
+
+// inspectRunningUpgradeJobs inspects all the jobs running in the cluster
+// and returns if the job updating the resource is already available
```
vharsh:
```suggestion
// and returns if even one of the the jobs updating the resource is already scheduled/running
```

pkg/upgrade/cstor.go
Abhishek-kumar09: Modified.

pkg/upgrade/cstor.go
Abhishek-kumar09: Done

pkg/upgrade/upgrade.go
Abhishek-kumar09: Discussed on call. It is to find any job with the underlying PV upgrade in progress.

pkg/upgrade/cstor.go
Abhishek-kumar09: Modified.

pkg/upgrade/cstor.go
Abhishek-kumar09: Replaced with List operation as suggested earlier.

pkg/upgrade/cstor.go
Abhishek-kumar09: Thanks, that was needed. Done.

pkg/client/cstor.go
```diff
@@ -284,3 +293,16 @@ func (k K8sClient) GetCSPIs(cspiNames []string, labelselector string) (*cstorv1.
 		Items: list,
 	}, nil
 }
+
+func (k K8sClient) GetCSPCOperator() (*corev1.Pod, error) {
```
vharsh: This function isn't required, there is already a `GetPods` method which can be used  with `("openebs.io/component-name=cspc-operator", "", "")` arguments.

pkg/client/k8s.go
```diff
@@ -276,6 +276,28 @@ func (k K8sClient) GetSC(scName string) (*v1.StorageClass, error) {
 	return sc, nil
 }
 
+// GetScWithCasType returns storage classes with the given cas-types
+func (k K8sClient) GetScWithCasType(casType string) ([]v1.StorageClass, error) {
```
vharsh: this function seems to be unused

pkg/upgrade/upgrade.go
vharsh: Add the reason for using an anonymous function over something else. I can only seem to recall "how not using an anon function made the nested for loop conditions more complex with extra booleans and how one cannot use a single `break` to come out of the jobItems loop". 


After a few more days, I'll require some more time to recall the need for having written a code this way, so a code-comment here would help a lot.

pkg/client/cstor.go
Abhishek-kumar09: Okay, I thought it will be handy in future. Onto it.

pkg/client/k8s.go
Abhishek-kumar09: Removing it.

Abhishek-kumar09:
Changes done.

vharsh (approved):
lgtm

PR #137, opened 2021-12-23
Update headings for CVRs under CStor vol describe

### LINKED ISSUE #135: Update cstor volume describe replica details heading to accurately specifiy underlying properties in an easy-to-grok way
The term `Total` and `Free` aren't super indicative of the spaces used by actual data or snapshots or metadata or at the application level and it does pop eyeballs when `Total < Free`.

## The Output

```
root@test-pcl4004:~# kubectl openebs describe volume pvc-7d5944e9-XXXXXXXXXXXX

pvc-7d5944e9-XXXXXXXXXXXX Details :
-----------------
NAME            : XXXXX
ACCESS MODE     : ReadWriteOnce
CSI DRIVER      : cstor.csi.openebs.io
STORAGE CLASS   : XXXX
VOLUME PHASE    : Bound
VERSION         : 2.10.0
CSPC            : XXXXXXXX
SIZE            : 2.0GiB
STATUS          : Healthy
REPLICA COUNT   : 1

Portal Details :
------------------
IQN              :  iqn.2016-09.com.openebs.cstor:XXXX
VOLUME NAME      :  XXXX
TARGET NODE NAME :  ABCD
PORTAL           :  XXXXX:3260
TARGET IP        :  XXXXX

Replica Details :
-----------------
NAME                                                            TOTAL     USED       STATUS    AGE
XXXXXXXXXXXXXXXXXXXXXXXXXSDJHFSHGHGHXXXXXXXXXXXXXXX-YYYYYYYZZZ   33.6MiB   126.8MiB   Healthy   67d1h
```

In the above output example, with the above headings it sort of becomes imperative that, `Total = Free + Used`, so in practice `Used` shouldn't exceed `Total`(unless `Free` is negative, which is *absurd*), but here the above values `total` means `actual data written to disk` & `used` means `how much data got referenced`.

At the nuts and bolts level, **TOTAL** comes from `cvr.spec.capacity.Total` which is **USED** property at the zfs level and `used` comes from `cvr.spec.capacity.Used` which is the `logicalreferenced` property at the zfs level.

# Tl; dr: What needs to be done?

1. The underlying ZFS metric needs to be understood well enough(this isn't super straightforward)
2. The understanding of the above metric should be conveyed properly via simple English phrases/words, such as `space used by data with compression, metadata, etc` or something simple such as `actual size on disk`, `logical size written by tenants`, etc.

vharsh:
fixes: #135 

Uses better terms to denote, `Total` and `Free` under _cstor volume describe output_ so that users don't think of it like `Total=Used+Free`(and get confused when Total < Used) outright but rather understand that `cvr.spec.capacity.Total` is `zfs used` and `cvr.spec.capacity.Used` is `zfs logical referenced`, wherein the earlier `total` can include all descendants like snapshots and is the actual zeros and ones written on the disk post compression(including snapshots, maybe clones, etc), where-as the earlier `used` refers to the data which the application(s) have written to the pool(minus the compression).

Co-Authored-by: Mayank Patel (@mynktl) <mayank.patel@mayadata.io>

### DIFF
--- a/pkg/util/constant.go
+++ b/pkg/util/constant.go
@@ -148,8 +148,8 @@ var (
 	// CstorReplicaColumnDefinations stores the Table headers for CVR Details
 	CstorReplicaColumnDefinations = []metav1.TableColumnDefinition{
 		{Name: "Name", Type: "string"},
-		{Name: "Total", Type: "string"},
-		{Name: "Used", Type: "string"},
+		{Name: "ZFS Used(compressed)", Type: "string"},
+		{Name: "LogicalReferenced", Type: "string"},
 		{Name: "Status", Type: "string"},
 		{Name: "Age", Type: "string"},
 	}


pkg/util/constant.go
```diff
@@ -148,8 +148,8 @@ var (
 	// CstorReplicaColumnDefinations stores the Table headers for CVR Details
 	CstorReplicaColumnDefinations = []metav1.TableColumnDefinition{
 		{Name: "Name", Type: "string"},
-		{Name: "Total", Type: "string"},
-		{Name: "Used", Type: "string"},
+		{Name: "ZFS Used(compressed)", Type: "string"},
+		{Name: "LogicalReferenced", Type: "string"},
```
vharsh, line 152: These two headings are suggestions for [this output](https://github.com/openebs/openebsctl/issues/135) please suggest better words which are concise but self-explanatory.

vharsh, line 152: These two headings are suggestions for [this output](https://github.com/openebs/openebsctl/issues/135) please suggest better words which are concise but self-explanatory.

mittachaitu (approved):
LGTM

PR #145, opened 2022-02-15
Fix GolangCI Lint missing errors

### LINKED ISSUE #105: Investigation: Not all tools of GolangCI lint are running in the CI

vharsh:
fixes: #105 

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,6 @@
 run:
-  timeout: 2m
+  deadline: 2m
+  issues-exit-code: 1
 
 linters-settings:
   goheader:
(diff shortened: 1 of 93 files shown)


niladrih (approved):
GolangCI job changes look good.

